PR #8, opened 2019-07-16
Fix unwanted links for Set [skip ci]

esparta:
The word Set was linked to the `Set` class.


esparta:
Fixed on 5be7da6f0f952539a75e89cae8c876ba6af9a429, @kou 

Thanks!

PR #9, opened 2019-07-29
Add a test for when ^ should fail to match

jneen:
Based on the discussion at https://github.com/ruby/strscan/pull/6#issuecomment-515832179, this tests that the behavior of `^` correctly fails to match when the stream is not at the beginning of a line.


kou:
Please don't reuse existing test...
https://github.com/ruby/strscan/pull/6#issuecomment-515847624
> But please add a new test method instead of changing existing one.

I'll fix and merge this later.

jneen:
It feels like the same test to me - testing the positive and negative aspects of the same thing - but I defer to you.

PR #17, opened 2020-12-17
Make strscan Ractor safe

mrkn:


mrkn:
Currently, macOS runners are unstable due to maintenance, so I ignore them.

PR #18, opened 2021-01-21
Replace "iff" with "if and only if"

gmcgibbon:
iff means if and only if, but readers without that knowledge might
assume this to be a spelling mistake. To me, this seems like
exclusionary language that is unnecessary. Simply using "if and only if"
instead should suffice.

strscan changes from https://github.com/ruby/ruby/pull/4035


kou:
Thanks.

PR #19, opened 2021-02-24
Use test-unit gem for test

hsbt:
Same as https://github.com/ruby/matrix/pull/25


olleolleolle (approved):
👍 This is the same kind of change, which was used in other places.

The usage of the sync_tool task was seen clearly in the final commit. Very neat.

Thanks for taking the time to update this. 

Also, moving the development dependencies to the Gemfile offer flexibility for the future.

PR #31, opened 2022-02-25
Remove non-US-ASCII characters from C source code

jeremyevans:
Replace characters with \u escapes.

No other C source files shipped with ruby include non-US-ASCII
characters.


kou:
Good catch!

PR #32, opened 2022-03-25
Document the need to require strscan library when using StringScanner

dafyddcrosby:
Tripped me up when using it for the first time, and looks like I'm not the first to hit this https://www.ruby-forum.com/t/strange-stringscanner-behaviour/50892

Thanks!

(originally posted on https://github.com/ruby/ruby/pull/5712)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ Or install it yourself as:
 ## Usage
 
 ```ruby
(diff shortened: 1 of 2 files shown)


kou:
Thanks!

Sorry. I missed this.

PR #33, opened 2022-04-05
Add Ruby 3.1 to the CI matrix

petergoldstein:
This PR runs green on my fork.  Just adds Ruby 3.1 to the CI matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
         - macos-10.15
         - windows-latest
         ruby:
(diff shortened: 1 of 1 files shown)


.ruby-version
```diff
@@ -0,0 +1 @@
+3.1.1
```
kou, line 1: Why do we need this?

petergoldstein: We didn't.  Errant check in.  I just fixed it.

kou:
Thanks.

PR #36, opened 2022-05-03
Simplify condition in strscan.gemspec

eregon:
From https://github.com/ruby/strscan/pull/25#discussion_r862348343


headius:
Aha, I just remembered why this exists: so that you can build the `java` gem with non-JRuby by forcing the RubyGems platform.

So it is up to @kou if this should merge or not. It would make it necessary to use JRuby to build the gem.

kou:
Thanks!

We don't need the condition because we build a gem for JRuby in CI.

PR #39, opened 2022-05-10
Move JRuby's ext and loader script

### LINKED ISSUE #38: strscan.rb move to causes two copies in JRuby stdlib
JRuby needs strscan.rb to be in the lib dir once installed, so that it can load the jar and extension. In d374f962e0e137d5d4abaa2d63ebeab6beede6a0 this file was moved to lib/jruby/strscan.rb to avoid conflicts with implementations that have their own strscan.rb. This fixes that problem, but by having two nested require paths in the gem we end up with the strscan.rb file being copied twice to the JRuby stdlib, with a confusing lib/jruby that should not exist.

The PR in #39 will move this JRuby-specific file into the ext dir and make appropriate adjustments elsewhere.

headius:
The strscan.rb file is only needed by JRuby, which does not have an extension file format that loads itself. As noted in #35 having this file present on TruffleRuby prevents their built-in strscan.rb from loading, so the file was moved to lib/jruby/strscan.rb with lib/jruby added as a require path. However this results in both lib/strscan.rb and lib/jruby/strscan.rb getting copied into JRuby's stdlib when the default gem is installed, leading to the confusing situation in #38.

This PR moves the jruby-specific strscan.rb under the jruby extension, renaming that directory to be explicitly for JRuby, and adding that new ext/jruby/lib path to the require paths so there's no nesting of require paths to copy files multiple times.

* ext/java => ext/jruby
* lib/jruby/strscan.rb => ext/jruby/lib/strscan.rb
* Changes to Rakefile and gemspec for new locations

Fixes #38


eregon:
Looks good to me :+1:

headius:
I believe this is ready. The only remaining failure is the known issue on Windows (which @enebo and I have discussed a bit... we are not encoding the error message correctly and will fix this later).

There's no urgent need to re-release with this change (nothing is actually broken right now), but it would be nice because Git shows me that there's an unversioned, unignored `lib/ruby/stdlib/jruby/strscan.rb` file that annoys me. 😀

kou:
Thanks!
Should we release a new version?

headius:
@kou That would be great, thank you!

kou:
Done!

PR #41, opened 2022-05-12
Include license in gem

marvinthepa:
why?

```
1. Redistributions of source code must retain the above copyright
   notice, this list of conditions and the following disclaimer.
2. Redistributions in binary form must reproduce the above copyright
   notice, this list of conditions and the following disclaimer in the
   documentation and/or other materials provided with the distribution.
```

AFAIK, a gem is a redistribution of source code and/or binary formats, so it should include the license. Also makes it easier for users of the gem to compile a list of all licenses they use.

### DIFF
--- a/strscan.gemspec
+++ b/strscan.gemspec
@@ -18,11 +18,11 @@ Gem::Specification.new do |s|
 
   if RUBY_ENGINE == 'jruby'
     s.require_paths = %w{ext/jruby/lib lib}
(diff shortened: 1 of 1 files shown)


kou:
Good catch!
Could you also add `COPYING`?

marvinthepa:
> Could you also add `COPYING`?

With the greatest pleasure.

kou:
OK. I'll merge this as is.

kou:
Thanks.

PR #47, opened 2022-07-13
Add dependabot to update GH actions

olleolleolle:
This only makes PRs with updates to the GitHub Actions versions. (Not gems, not any other packages.)


kou:
Thanks! Let's try this configuration.

PR #49, opened 2023-01-29
Mention return value of `rest?` in the doc

okuramasafumi:
The doc of `rest?` was unclear about return value. This commit adds the return value to the doc.


kou:
Thanks!

Issue #50, opened 2023-02-02
Two helpful methods: currchar and nextchar


shreeve:
Edited: This was the original message, which has been changed since then.

I'm working on a small but very performant CSV parser and am using `strscan` in several key places. The `strscan` library is really amazing and the code is beautiful, but I believe the addition of 3 methods would be extremely helpful. Of course, I can and have mimicked these methods in my Ruby code, but adding them to `strscan` directly would be a performance boost.

The methods are:

1. `scan_upto`: This would work identically to `scan_until`, except it would *not* include the match. Currently, if my string is `abcdef` and I run `s.scan_until(/de/)`, it will return `abcde` and the pointer will be positioned at `f`. In contrast, `s.scan_upto(/de/)` would return `abc` and the pointer would be positioned at `d` (ie - everything "up to, but not including" what I scanned for).

2. `next_char`: Currently, we can use `getch` to return the "current" character and then advance the pointer. So, if I have a scanner with `abcdef` and `s.pos` is 3 (ie - on the `d` character), then when I call `getch`, it will return the `d` but advance the pointer to `e`. I then have to go back and read again to get the `e` if I want to know the "next" character. If I run `s.peek(1)` I can get the `e`, but `peek()` is not multibyte-aware, so I may only get a byte of a multi-byte character. I propose that the `next_char` method would work pretty much like `getch` does today, except it would first advance the pointer one *character* (in a multibyte-aware manner, just like `getch` does today), but then it would return the now-current character (again, multibyte aware) at the next position. In the example mentioned, if I have `abcdef` and am on the `d`, then `next_char` will advance to the `e` and return the `e`.

3. `curr_char`: Right now, I can call `s.string[s.pos]` to get the character at the current position. But, for some reason, this method isn't as fast as I would like it to be. When trying to write highly performant code, this slowdown is noticeable. Instead, I would like for `curr_char` to simply do the same thing as `getch` does today (and is multibyte aware), except *not* advance the pointer. Right now, I've been using `s.peek(1)` to read the current character (it's faster than `s.string[s.pos]`), but `s.peek(1)` is not multibyte aware, so it's not safe to use, in the general sense.

These three small methods would provide a very welcome little "boost" for some code that I am working on and have had to do in Ruby code until now.

shreeve:
The above code *does not* use these three new methods, but would benefit from them.

shreeve:
Going to close this and create a new, simpler issue.

PR #51, opened 2023-02-02
Add three new methods, see https://github.com/ruby/strscan/issues/50

shreeve:
This PR implements the three methods (`scan_upto`, `next_char`, and `curr_char`) as discussed in:

https://github.com/ruby/strscan/issues/50


shreeve:
Duplicate

PR #52, opened 2023-02-02
Add three methods

shreeve:
See issue https://github.com/ruby/strscan/issues/50

### DIFF
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -110,6 +110,7 @@ static VALUE strscan_check_until _((VALUE self, VALUE re));
 static VALUE strscan_search_full _((VALUE self, VALUE re,
                                     VALUE succp, VALUE getp));
 static void adjust_registers_to_matched _((struct strscanner *p));
+static VALUE strscan_curr_char _((VALUE self));
 static VALUE strscan_getch _((VALUE self));
 static VALUE strscan_get_byte _((VALUE self));
 static VALUE strscan_getbyte _((VALUE self));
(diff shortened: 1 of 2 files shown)


ext/strscan/strscan.c
```diff
@@ -620,10 +623,16 @@ strscan_do_scan(VALUE self, VALUE pattern, int succptr, int getstr, int headonly
...
         succ(p);
+        if (succptr == 2) p->curr = p->regs.beg[0];
     }
     {
         const long length = last_match_length(p);
         if (getstr) {
+            if (getstr == 2) {
+                return extract_range(p,
+                                     0,
+                                     adjust_register_position(p, p->regs.beg[0]));
```
shreeve, line 634: This is not correct, but I'm not familiar enough with the code to know how to change this. Essentially, we just want to return "up to" (but not including) the matched regex.

shreeve:
This PR originally had three new methods:

* `curr_char`
* `next_char`
* `scan_upto`

These have been changed as follows:

* `curr_char` has become `peekch` - To stay in line with the concept of "peeking" by looking, but not advancing, this `peekch` allows you to peek forward one *character* (not one *byte*). As such, it is now multibyte aware. Unlike `peek`, this method doesn't take an argument now, and just returns the current (multibyte aware) character at the current scan position.
* `next_char` has become `nextch` - This work almost the same as `getch`, except `getch` returns the character at the current position and then advances the pointer. This `nextch` method first advances the pointer one character and then returns the character at that (now the current) position.
* `scan_upto` was determined to not be needed because you can always do a positive lookahead to accomplish the same this. In `scan_until`, if you have "Dec 12 2022" and you are at the "e" and you use `s.scan_until(/20/)` you will get back "ec 12 20". But, if you use `s.scan_until(/(?=20)/)`, then you will get back "ec 12 ". Since this is already doable, I removed the code I had for `scan_upto`.

Does it look like the `peekch` and `nextch` methods could be added?

shreeve:
The names `peekch` doesn't really work:

* `peekch` - Two problems here, first: unlike `peek` we want to return `nil` (unlike `peek`, which returns an empty string "") if we are at the end of the input. Second, `peek` takes a length parameter and `peekch` doesn't, so the name isn't really a fit.

For now, I'll close this PR and will see if there's an even simpler approach.

PR #55, opened 2023-02-06
Introduce an adapter from joni 2.1 to 2.2

headius:
JOni 2.2 will hide these public fields and the direct constructors for Region, so these adapters will bridge the gap until the new API is widely available. The adapter and calls to it can be removed once JRuby has shipped a few releases of 9.4 (and possibly 9.3) with the updated JOni.

### DIFF
--- a/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
+++ b/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
@@ -27,6 +27,7 @@
 
 package org.jruby.ext.strscan;
 
+import jnr.a64asm.REG;
 import org.jcodings.Encoding;
 import org.joni.Matcher;
 import org.joni.NameEntry;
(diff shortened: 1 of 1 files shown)


headius:
See https://github.com/jruby/jruby/pull/7620 which moves JRuby to the new API, https://github.com/jruby/joni/pull/58 and https://github.com/jruby/joni/pull/60 which introduce the API into joni, and https://github.com/jruby/joni/pull/59 which will remove access to the public fields and constructors in joni 2.2.

headius:
This will not build cleanly until JRuby is released with the updated joni API. Merge and release after JRuby 9.4.1.

kou (approved):
+1

You can merge this after JRuby 9.4.1 release.

headius:
This now builds cleanly and only fails the one known failure on Windows. Merging!

headius:
@kou This is ready for a release any time!

kou:
Done!

Issue #56, opened 2023-02-08
Some tests not running in repository


headius:
I discovered today that some of the tests in the repository do not seem to be running. Specifically it seems like the inherited tests for TestStringScannerFixedAnchor are omitted.

This masked the fact that JRuby did not properly implement `fixed_anchor`, which I'm now fixing.

Here's the verbose test output from command line, showing that only four tests run with `fixed_anchor` behavior enabled:

```
[] strscan $ jruby -I lib -I ext/jruby/lib -I test/lib -rhelper test/strscan/test_stringscanner.rb -v
/Users/headius/.gem/jruby/3.1.0/gems/power_assert-2.0.3/lib/power_assert.rb:8: warning: tracing (e.g. set_trace_func) will not capture all events without --debug flag
/Users/headius/.gem/jruby/3.1.0/gems/power_assert-2.0.3/lib/power_assert.rb:8: warning: tracing (e.g. set_trace_func) will not capture all events without --debug flag
Loaded suite test/strscan/test_stringscanner
Started
TestStringScanner: 
  test_AREF:																						.: (0.020898)
  test_aref_without_regex:																				.: (0.002298)
  test_bol?:																						.: (0.012420)
  test_captures:																					.: (0.001524)
  test_charpos_not_use_string_methods:																			.: (0.008971)
  test_check:																						.: (0.002447)
  test_check_until:																					.: (0.006086)
  test_concat:																						.: (0.004053)
  test_const_Id:																					.: (0.006021)
  test_const_Version:																					.: (0.001307)
  test_dup:																						.: (0.006789)
  test_encoding:																					.: (0.001849)
  test_encoding_string:																					.: (0.000913)
  test_eos?:																						.: (0.002434)
  test_exist_p:																						.: (0.001809)
  test_exist_p_string:																					.: (0.002137)
  test_fixed_anchor_false:																				.: (0.006418)
  test_fixed_anchor_true:																				.: (0.004086)
  test_generic_regexp:																					.: (0.001350)
  test_get_byte:																					.: (0.002466)
  test_getch:																						.: (0.003643)
  test_inspect:																						.: (0.002125)
  test_inspect2:																					.: (0.002130)
  test_invalid_encoding_string:																				.: (0.021611)
  test_match_p:																						.: (0.001632)
  test_matched:																						.: (0.005154)
  test_matched_size:																					.: (0.002953)
  test_named_captures:																					.: (0.002098)
  test_peek:																						.: (0.002760)
[...]

PR #57, opened 2023-02-08
Finish fixed_anchor support in the JRuby ext

headius:
This support was never finished, causing several tests in TestStringScannerFixedAnchor to fail. However, because of #56 I did not realize these were still failing.

@kou This can be merged and released any time, and I have confirmed in the JRuby repository that it does indeed allow all remaining `fixed_anchor` tests to pass. I have no explanation for #56 right now.


headius:
I am also looking into that one failure on Windows. I believe it is a bug in our regular expression engine, joni, where it improperly decodes the group name for the error message.

headius:
Looks good on local runs, #56 notwithstanding. Merging.

PR #58, opened 2023-02-08
Add test case to `test_string`

okuramasafumi:
`string` returns the original string after `scan` is called. Current test doesn't check this behavior and now it's covered.


kou:
Thanks!

PR #61, opened 2023-02-11
Fix a crash when `named_captures` called on new instance

okuramasafumi:
```ruby
StringScanner.new('foo').named_captures
```

This commit fixes a crash in the case above.

### DIFF
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -1502,7 +1502,9 @@ strscan_named_captures(VALUE self)
     named_captures_data data;
     data.self = self;
     data.captures = rb_hash_new();
-    onig_foreach_name(RREGEXP_PTR(p->regex), named_captures_iter, &data);
+    if (!RB_NIL_P(p->regex)) {
+        onig_foreach_name(RREGEXP_PTR(p->regex), named_captures_iter, &data);
+    }
 
     return data.captures;
 }
--- a/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
+++ b/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
@@ -829,6 +829,8 @@ public class RubyStringScanner extends RubyObject {
 
         RubyHash captures = RubyHash.newHash(runtime);
 
+        if (pattern == null) return captures;
+
         Iterator<NameEntry> nameEntryIterator = pattern.namedBackrefIterator();
 
         while (nameEntryIterator.hasNext()) {
--- a/test/strscan/test_stringscanner.rb
+++ b/test/strscan/test_stringscanner.rb
@@ -775,6 +775,7 @@ module StringScannerTests
   def test_named_captures
     omit("not implemented on TruffleRuby") if ["truffleruby"].include?(RUBY_ENGINE)
     scan = StringScanner.new("foobarbaz")
+    assert_equal({}, scan.named_captures)
     assert_equal(9, scan.match?(/(?<f>foo)(?<r>bar)(?<z>baz)/))
     assert_equal({"f" => "foo", "r" => "bar", "z" => "baz"}, scan.named_captures)
   end


okuramasafumi:
CI fails since a new test case for crash was added.

ext/strscan/strscan.c
```diff
@@ -1502,7 +1502,9 @@ strscan_named_captures(VALUE self)
     named_captures_data data;
     data.self = self;
     data.captures = rb_hash_new();
-    onig_foreach_name(RREGEXP_PTR(p->regex), named_captures_iter, &data);
+    if (p->regex != Qnil) {
```
kou, line 1505:
```suggestion
    if (!RB_NIL_P(p->regex)) {
```

kou (commented):
Could you also add a `NULL` check to the implementation.

The CI job failures show that installed strscan gem isn't used (Ruby's strscan is used)...

okuramasafumi:
@kou Thank you, I've fixed it.
https://github.com/ruby/strscan/blob/master/.github/workflows/ci.yml#L62
This line specifies Ruby version as 3.1, but should we add 3.2 to avoid CI failures?

kou:
> Could you also add a `NULL` check to the implementation.

Oh, sorry. I wanted to say "Could you also add a `NULL` check to the JRuby implementation?"

> https://github.com/ruby/strscan/blob/master/.github/workflows/ci.yml#L62
> This line specifies Ruby version as 3.1, but should we add 3.2 to avoid CI failures?

It's better but it's not related to the CI failures.

okuramasafumi:
@kou OK, so I changed JRuby version as well and `run-test` to use locally installed version. I'm not sure if these fixes are ideal but CI now passes. Could you check it, thanks!

run-test.rb
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env ruby
 
+require "bundler/setup"
```
kou, line 3: If we add this, we always use local strscan not strscan gem.

https://github.com/ruby/strscan/actions/runs/4160453753/jobs/7197443977#step:11:5

```text
Loaded strscan from /home/runner/work/strscan/strscan/lib/strscan.so
Gem from /home/runner/work/strscan/strscan
```

We should use strscan gem like https://github.com/ruby/strscan/actions/runs/4134931150/jobs/7146703286#step:11:5 :

```text
Loaded strscan from /opt/hostedtoolcache/Ruby/3.1.3/x64/lib/ruby/gems/3.1.0/gems/strscan-3.0.7/lib/strscan.so
Gem from /opt/hostedtoolcache/Ruby/3.1.3/x64/lib/ruby/gems/3.1.0/gems/strscan-3.0.7
```

okuramasafumi: I see, I reverted that change with force push.
Should I do something to make CI pass? I believe after releasing a new gem including this fix Ci will become green.
Or more general question, what should we do if a fix and a test are included in the same commit?

kou: > Should I do something to make CI pass?

Can you take a look at why installed gem isn't used only with Ruby 3.2 and master?
Or could you create an issue for this?

It may be a RubyGems/Bundler bug in Ruby 3.2 and master... 

> I believe after releasing a new gem including this fix Ci will become green.

It's not the intention for the test. The test should use "not-released" gem to prevent releasing a buggy gem.

>  Or more general question, what should we do if a fix and a test are included in the same commit?

If our CI jobs work as expected (installed gem is used), our test will be passed in the case.

okuramasafumi: OK so this might be an issue outside of strscan, now I understand.
I'll try to resolve it myself, and if it's too difficult I'll submit an issue.

okuramasafumi:
The output from `gem install --verbose --backtrace pkg/*.gem`.

`3.0`
https://github.com/ruby/strscan/actions/runs/4171325802/jobs/7221150111
> ["/opt/hostedtoolcache/Ruby/3.0.5/x64/bin/ruby", "-I", "/opt/hostedtoolcache/Ruby/3.0.5/x64/lib/ruby/site_ruby/3.0.0", "extconf.rb"]

`3.1`
https://github.com/ruby/strscan/actions/runs/4171325802/jobs/7221150216
> ["/opt/hostedtoolcache/Ruby/3.1.3/x64/bin/ruby", "-I", "/opt/hostedtoolcache/Ruby/3.1.3/x64/lib/ruby/site_ruby/3.1.0", "extconf.rb"]

`3.2`
https://github.com/ruby/strscan/actions/runs/4171325802/jobs/7221150323
> ["/opt/hostedtoolcache/Ruby/3.2.1/x64/bin/ruby", "-I", "/opt/hostedtoolcache/Ruby/3.2.1/x64/lib/ruby/3.2.0", "extconf.rb"]

It looks like `3.2` include path is different from `3.1` and `3.0` that are successful. This is might be related to this problem.

tenderlove:
What do we need to do to merge this?  I ran in to the same bug. 😅

kou:
We need to debug a problem that Ruby 3.2 and Ruby master doesn't use `gem install`-ed strscan.
But we can work on the problem as a separated pull request and merge this as-is.

hsbt:
FYI: I fixed installation failure related C-ext default gems https://github.com/rubygems/rubygems/pull/6430.

I'm not sure this fix resolve issue of this PR.

kou:
Oh, thanks for the info.
Can we try the change in this repository?

kou:
I merge this for now.
We can work on the CI failures in a separated pull request.

eregon:
> I merge this for now. We can work on the CI failures in a separated pull request.

How was this fixed? I don't see any change in https://github.com/ruby/strscan/commits/master/.github/workflows
Maybe simply by updating to newer CRuby versions which have that RubyGems fix in?

PR #63, opened 2023-03-06
Fix crash when no scanning has been done

tenderlove:
`StringScanner.new("foo").named_captures` will crash because no regular expression has been set.  This commit returns an empty hash if no re is set on the string scanner.

We could probably also return `nil`, but I'm not sure.  I think an empty hash is probably right since most callers expect a hash to be returned.


tenderlove:
lol oops.  I found the same crash as #61

PR #64, opened 2023-07-27
Use new function rb_reg_onig_match

peterzhu2118:
rb_reg_onig_match performs preparation, error handling, and cleanup for matching a regex against a string. This reduces repetitive code and removes the need for StringScanner to access internal data of regex.
    
This commit syncs strscan to commit ruby/ruby@7193b404a1a56e50f8046d0382914907020c1559.

### DIFF
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -539,6 +539,68 @@ adjust_register_position(struct strscanner *p, long position)
     }
 }
 
+/* rb_reg_onig_match is available in Ruby 3.3 and later. */
+#ifndef HAVE_RB_REG_ONIG_MATCH
+static OnigPosition
+rb_reg_onig_match(VALUE re, VALUE str,
+                  OnigPosition (*match)(regex_t *reg, VALUE str, struct re_registers *regs, void *args),
+                  void *args, struct re_registers *regs)
+{
+    regex_t *reg = rb_reg_prepare_re(re, str);
+
+    bool tmpreg = reg != RREGEXP_PTR(re);
+    if (!tmpreg) RREGEXP(re)->usecnt++;
+
+    OnigPosition result = match(reg, str, regs, args);
+
+    if (!tmpreg) RREGEXP(re)->usecnt--;
+    if (tmpreg) {
+        if (RREGEXP(re)->usecnt) {
+            onig_free(reg);
+        }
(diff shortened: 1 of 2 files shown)


ext/strscan/strscan.c
```diff
@@ -539,6 +539,68 @@ adjust_register_position(struct strscanner *p, long position)
...
+strscan_match(regex_t *reg, VALUE str, struct re_registers *regs, void *args_ptr)
+{
+    struct strscanner *p = (struct strscanner *)args_ptr;
+
+    return onig_match(reg,
+               match_target(p),
+               (UChar* )(CURPTR(p) + S_RESTLEN(p)),
+               (UChar* )CURPTR(p),
+               regs,
+               ONIG_OPTION_NONE);
```
kou, line 587:
```suggestion
                      match_target(p),
                      (UChar* )(CURPTR(p) + S_RESTLEN(p)),
                      (UChar* )CURPTR(p),
                     regs,
                     ONIG_OPTION_NONE);
```

ext/strscan/strscan.c
```diff
@@ -539,6 +539,68 @@ adjust_register_position(struct strscanner *p, long position)
...
+        }
+        else {
+            onig_free(RREGEXP_PTR(re));
+            RREGEXP_PTR(re) = reg;
+        }
+    }
+
+    if (result < 0) {
+        if (result != ONIG_MISMATCH) {
+            rb_raise(ScanError, "regexp buffer overflow");
```
kou, line 569: Hmm. https://github.com/ruby/ruby/pull/8123/files#diff-c3675fa319803b2f5a775defa40694edb9a761baa3a54fa78e1fdef8f918cc7cR1652 is backward incompatible change.
But this will not be happen in general use?

kou (commented):
Could you write what is done by PR instead of "syncs strscan" for the PR title and description?
We use the PR title and description for commit message.

ext/strscan/strscan.c
peterzhu2118: Yeah I think this should never happen, and if it does, the new code will have better error handling.

peterzhu2118:
Thanks for the review @kou! I updated the commit and PR messages.

kou (approved):
+1

PR #65, opened 2023-07-28
Use test-unit-ruby-core instead of deep-copy from ruby repository

hsbt:
I and @nobu prepared https://github.com/ruby/test-unit-ruby-core for `test-unit` with `EnvUtil`.

We should remove related files from this repository.


hsbt:
I'm not sure why Ruby 3.2 crashed.

https://github.com/ruby/strscan/actions/runs/6584775176/job/17889936696

kou:
Because strscan in Ruby 3.2 not built strscan isn't used.
#61 isn't included in strscan in Ruby 3.2.

kou (approved):
The crash isn't related to this change.

hsbt:
👍  Thanks for merging this.

PR #67, opened 2023-10-10
Minor cleanup and optimization

headius:
A few minor tweaks.

* Cache field accesss throughout
* Arity-split `values_at` method for low arity calls
* General cleanup


headius:
@mrkn Could we do a release soon? These are small changes but also minor perf improvements and good cleanup.

kou:
I can do it.

kou:
Done.

Issue #70, opened 2024-01-05
Response when there is no substring matching the group in StringScanner#captures


naitoh:
If there is no substring matching the group (s[3]), the behavior is different.

If there is no substring matching the group, the corresponding element (s[3]) should be nil.

```
s = StringScanner.new('foobarbaz') #=> #<StringScanner 0/9 @ "fooba...">
s.scan /(foo)(bar)(BAZ)?/  #=> "foobar"
s[0]           #=> "foobar"
s[1]           #=> "foo"
s[2]           #=> "bar"
s[3]           #=> nil
s.captures #=> ["foo", "bar", ""]
s.captures.compact #=> ["foo", "bar", ""]
```

### FYI

https://docs.ruby-lang.org/ja/latest/method/MatchData/i/captures.html
```
/(foo)(bar)(BAZ)?/ =~ "foobarbaz" #=> 0
$~.to_a        #=> ["foobar", "foo", "bar", nil]
$~.captures #=> ["foo", "bar", nil]
$~.captures.compact #=> ["foo", "bar"]
```

* StringScanner#captures is not yet documented.
https://docs.ruby-lang.org/ja/latest/class/StringScanner.html

PR #72, opened 2024-01-08
StringScanner#captures: Return nil not "" for unmached capture

naitoh:
fix https://github.com/ruby/strscan/issues/70
If there is no substring matching the group (s[3]), the behavior is different.

If there is no substring matching the group, the corresponding element (s[3]) should be nil.

### Before
```
s = StringScanner.new('foobarbaz') #=> #<StringScanner 0/9 @ "fooba...">
s.scan /(foo)(bar)(BAZ)?/  #=> "foobar"
s[0]           #=> "foobar"
s[1]           #=> "foo"
s[2]           #=> "bar"
s[3]           #=> nil
s.captures #=> ["foo", "bar", ""]
s.captures.compact #=> ["foo", "bar", ""]
```

### After
```
s = StringScanner.new('foobarbaz') #=> #<StringScanner 0/9 @ "fooba...">
s.scan /(foo)(bar)(BAZ)?/  #=> "foobar"
s[0]           #=> "foobar"
s[1]           #=> "foo"
s[2]           #=> "bar"
s[3]           #=> nil
s.captures #=> ["foo", "bar", nil]
s.captures.compact #=> ["foo", "bar"]
```

### FYI

https://docs.ruby-lang.org/ja/latest/method/MatchData/i/captures.html
```
/(foo)(bar)(BAZ)?/ =~ "foobarbaz" #=> 0
$~.to_a        #=> ["foobar", "foo", "bar", nil]
$~.captures #=> ["foo", "bar", nil]
$~.captures.compact #=> ["foo", "bar"]
```

* StringScanner#captures is not yet documented.
https://docs.ruby-lang.org/ja/latest/class/StringScanner.html

### DIFF
--- a/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
+++ b/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
@@ -901,9 +901,14 @@ public class RubyStringScanner extends RubyObject {
         newAry = RubyArray.newArray(runtime, numRegs);
 
         for (i = 1; i < numRegs; i++) {
-            IRubyObject str = extractRange(runtime,
+            IRubyObject str;
+            if (REGION_ADAPTER.getBeg(regs, i) == -1) {
+                str = context.nil;
+            } else {
+                str = extractRange(runtime,
                     adjustRegisterPosition(REGION_ADAPTER.getBeg(regs, i)),
                     adjustRegisterPosition(REGION_ADAPTER.getEnd(regs, i)));
+            }
             newAry.push(str);
         }
 
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -1243,10 +1243,10 @@ strscan_size(VALUE self)
  * If nothing was priorly matched, it returns nil.
  *
  *   s = StringScanner.new("Fri Dec 12 1975 14:39")
- *   s.scan(/(\w+) (\w+) (\d+) /)       # -> "Fri Dec 12 "
- *   s.captures                         # -> ["Fri", "Dec", "12"]
- *   s.scan(/(\w+) (\w+) (\d+) /)       # -> nil
- *   s.captures                         # -> nil
+ *   s.scan(/(\w+) (\w+) (\d+) (1980)?/)       # -> "Fri Dec 12 "
+ *   s.captures                                # -> ["Fri", "Dec", "12", nil]
+ *   s.scan(/(\w+) (\w+) (\d+) (1980)?/)       # -> nil
+ *   s.captures                                # -> nil
  */
 static VALUE
 strscan_captures(VALUE self)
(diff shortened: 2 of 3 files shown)


ext/strscan/strscan.c
```diff
@@ -1262,9 +1263,12 @@ strscan_captures(VALUE self)
     new_ary  = rb_ary_new2(num_regs);
 
     for (i = 1; i < num_regs; i++) {
-        VALUE str = extract_range(p,
-                                  adjust_register_position(p, p->regs.beg[i]),
-                                  adjust_register_position(p, p->regs.end[i]));
+        if (p->regs.beg[i] == -1)
```
kou, line 1266: Could you declare this variable in this scope to narrow scope as much as possible?

```suggestion
        VALUE str;
        if (p->regs.beg[i] == -1)
```

ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
```diff
@@ -901,9 +902,13 @@ public IRubyObject captures(ThreadContext context) {
         newAry = RubyArray.newArray(runtime, numRegs);
 
         for (i = 1; i < numRegs; i++) {
-            IRubyObject str = extractRange(runtime,
```
kou, line 904: Could you declare this variable here?

ext/strscan/strscan.c
naitoh: OK.
I See.

ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
naitoh: OK.
I See.

ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
```diff
@@ -901,9 +901,14 @@ public IRubyObject captures(ThreadContext context) {
...
+            IRubyObject str;
+            if (REGION_ADAPTER.getBeg(regs, i) == -1) {
+                str = context.nil;
+            } else {
+                str = extractRange(runtime,
                     adjustRegisterPosition(REGION_ADAPTER.getBeg(regs, i)),
                     adjustRegisterPosition(REGION_ADAPTER.getEnd(regs, i)));
```
kou, line 910: @headius Should we indent them?

kou: I'll merge this. If we need to fix style here, we can do it as a follow-up PR.

headius: We generally just try to match existing style when making other changes. We can do a separate PR to clean up formatting.

headius:
And thank you for patching the JRuby code!

Issue #73, opened 2024-01-08
CI: perhaps we need Ruby 3.3 in the matrix


olleolleolle:
The build matrix does not yet include all the versions. 

Perhaps using the ruby GitHub Action that fetches versiona externally is the way to get there easiest and most future-cheap.

PR #74, opened 2024-01-08
CI: Use shared versions GH Actions workflow

olleolleolle:
fix ruby/strscan#73

I took some inspiration from https://github.com/ruby/uri/blob/master/.github/workflows/test.yml and then I edited the necessary values.

(I put this as a Draft PR, since I am not sure about all the details, which engines and versions will be included.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
 
   host:
     needs: ruby-versions
-    name: build (${{ matrix.ruby }} / ${{ matrix.os }})
+    name: ${{ matrix.os }} ${{ matrix.ruby }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,35 +5,28 @@ on:
 - pull_request
 
 jobs:
+  ruby-versions:
+    uses: ruby/actions/.github/workflows/ruby_versions.yml@master
+    with:
+      min_version: 2.4
+
   host:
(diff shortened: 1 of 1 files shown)


eregon:
FYI, CI fails on jruby latest release, tracked as https://github.com/ruby/strscan/issues/75

.github/workflows/ci.yml
```diff
@@ -5,35 +5,28 @@ on:
...
 jobs:
+  ruby-versions:
+    uses: ruby/actions/.github/workflows/ruby_versions.yml@master
+    with:
+      min_version: 2.4
+
   host:
-    name: ${{ matrix.os }} ${{ matrix.ruby }}
+    needs: ruby-versions
+    name: build (${{ matrix.ruby }} / ${{ matrix.os }})
```
kou, line 15: Can we keep the original `name:` for short label?

kou (commented):
@headius Do we need both of `jruby` and `jruby-head`?

.github/workflows/ci.yml
olleolleolle: Yes, now done!

olleolleolle:
I learned about the JRuby detail:

> Yeah we had a transition period where it supported both old and new but now all released JRuby should be on **the new version**.

And by "the new version" headius is referring to the Joni library version.

headius:
I have merged #76 and master is now green on latest JRuby.

headius:
@olleolleolle Thanks for the PR!

olleolleolle:
In #82, kou explained the reason for the failed builds on current master (and also in this PR).

I will merge this, now.

PR #76, opened 2024-01-10
JRuby extension: drop old-Joni-version support

olleolleolle:
This PR removes the code path for OldRegionAdapter, now that all published JRuby versions are using "new API" for `Region`, so no adapting is necessary.

## Details

headius mentioned in an off-GitHub conversation:
> Yeah we had a transition period where it supported both old and new
but now all released JRuby should be on the new version

> (Using Region) should be fine, especially since 9.3 can't really install the gem (it will be ignored unless explicitly activated) and 9.4 has had the new region logic for several releases

## TODO

- [x] Remove the Interface, and use Region as-is.
- [x] Remove the usage of the intermediate interface

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
           - debug
           # We need to update 'jruby-XXX' in "if:" too when we update
           # this.
-          - jruby-9.4.1.0
+          - jruby-9.4.5.0
           - truffleruby
           - truffleruby-head
         include:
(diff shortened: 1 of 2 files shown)


headius:
Expanding my reply to @olleolleolle:

* JRuby versions prior to JRuby 9.4 will not use the gem if it is not explicitly activated, since they bundled it as a non-gem part of stdlib.
* JRuby 9.2 and earlier will break if this gem is upgraded and activated, but we believe there are few applications still using 9.2 that could not easily be upgraded to 9.3 or 9.4.
* JRuby 9.3.11.0 and newer all ship with the updated joni logic and can support the gem without the legacy code.

I think this is fine to do. It's not a commonly upgraded gem, and even if it does get upgraded, JRuby has supported these APIs for several months.

@enebo Any concerns?

headius:
Expanding my reply to @olleolleolle:

* JRuby versions prior to JRuby 9.4 will not use the gem if it is not explicitly activated, since they bundled it as a non-gem part of stdlib.
* JRuby 9.2 and earlier will break if this gem is upgraded and activated, but we believe there are few applications still using 9.2 that could not easily be upgraded to 9.3 or 9.4.
* JRuby 9.3.11.0 and newer all ship with the updated joni logic and can support the gem without the legacy code.

I think this is fine to do. It's not a commonly upgraded gem, and even if it does get upgraded, JRuby has supported these APIs for several months.

@enebo Any concerns?

headius (approved):
Looks good!

PR #87, opened 2024-02-03
Don't add begin to length for new string slice

### LINKED ISSUE #86: JRuby ext overflows slicing strings with non-zero begin
When the string provided to `StringScanner.new` is represented by a ByteList with a non-zero `begin` offset, the length for new strings is calculated incorrectly (length and begin get added together). No tests exist for this case, which typically happens when the source string has itself been sliced off the end of a larger string and still shares the same buffer.

```
$ jruby -rstrscan -e 'str = "hellohello"[5..-1]; StringScanner.new(str).scan(/hello/)'
Unhandled Java exception: java.lang.ArrayIndexOutOfBoundsException: arraycopy: last source index 15 out of bounds for byte[10]
java.lang.ArrayIndexOutOfBoundsException: arraycopy: last source index 15 out of bounds for byte[10]
            arraycopy at java/lang/System.java:-2
               <init> at org/jruby/util/ByteList.java:227
            newString at org/jruby/ext/strscan/RubyStringScanner.java:877
        extractBegLen at org/jruby/ext/strscan/RubyStringScanner.java:260
                 scan at org/jruby/ext/strscan/RubyStringScanner.java:342
                 scan at org/jruby/ext/strscan/RubyStringScanner.java:391
```

This is the cause of jruby/jruby#8083. FIx coming.

headius:
Fixes #86

### DIFF
--- a/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
+++ b/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
@@ -874,7 +874,7 @@ public class RubyStringScanner extends RubyObject {
         ByteList byteList = str.getByteList();
         int begin = byteList.begin();
 
-        ByteList newByteList = new ByteList(byteList.unsafeBytes(), begin + start, begin + length, byteList.getEncoding(), true);
+        ByteList newByteList = new ByteList(byteList.unsafeBytes(), begin + start, length, byteList.getEncoding(), true);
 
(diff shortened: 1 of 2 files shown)


headius:
This can be released any time. Equivalent to a segv for JRuby so sooner is better. 😀

kou:
Done!

headius:
@kou Thank you!

PR #91, opened 2024-03-26
Omit tests for `#scan_byte` and `#peek_byte` on TruffleRuby temporary

andrykonchin:
The methods were added in #89 but they aren't implemented in TruffleRuby yet. So let's skip them for now to have CI green.


kou (approved):
+1

PR #93, opened 2024-05-03
Remove `dir-locals.el`

okuramasafumi:
This might have been added accidentally with a369e38c, but this file is in the list of gitignore by GitHub. https://github.com/github/gitignore/blob/main/Global/Emacs.gitignore#L45

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,1 +1,0 @@
-((c-mode . ((c-file-style . "ruby"))))


kou:
It's intentional because we use the Ruby style for our code base.
Should we remove this?

okuramasafumi:
Sorry I didn't notice it's intentional.
I think we can introduce more general way such as https://editorconfig.org/ to make editors work as expected.
If you'd agree, I'll add an editorconfig file with equivalent settings.

knu:
I don't think C-specific formatting rules that ruby-style.el defines can be implemented with .editorconfig.

Besides, the presence of this file does no harm to anyone and I see no point in bothering with what maintainers are comfortable with.  There's one in the ruby source tree, too. https://github.com/ruby/ruby/blob/master/.dir-locals.el

Tip: If you are an Emacs user and want to override it, put yours as .dir-locals-2.el.
https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html

kou:
If we want to use more general approach, how about using `clang-format` and `pre-commit` instead of editor configuration?

FYI: Here are configurations I used in other project:
* https://github.com/apache/arrow-flight-sql-postgresql/blob/main/.clang-format
* https://github.com/apache/arrow-flight-sql-postgresql/blob/main/.pre-commit-config.yaml

If we have them, we can format our code `pre-commit run -a` (and use it in CI). If we have a `clangd` integration such as https://bugs.ruby-lang.org/issues/20034 , we can use the configuration in an editor via LSP.

One challenge of it is that we need to create a `.clang-format` that is compatible with https://github.com/ruby/ruby/blob/master/misc/ruby-style.el ...

okuramasafumi:
Thank you for sharing info. I'm closing this now since it's basically my misunderstanding that it's added accidentally, and I'm not familiar with `clang-format` as an alternative solution.

PR #94, opened 2024-05-05
Exclude older than 2.6 on macos-14

nobu:


eregon (approved):
Indeed, Ruby < 2.6 is not available on darwin-aarch64 for setup-ruby because it failed to build: https://github.com/ruby/setup-ruby/pull/494#issuecomment-1758099885

PR #98, opened 2024-05-22
Update Java ext version and add script logic for bump

### LINKED ISSUE #97: StringScanner::Version is not getting updated for JRuby
It is currently hardcoded in RubyStringScanner.java here: https://github.com/ruby/strscan/blob/master/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java#L91

From what I can gather this version gets updated in the C code using Rake target `version:bump` but that only updates the C code.

headius:
The Java version is hardcoded like the C version, but has not been getting updated. This will at least let it be updated by the rake target version:bump, but any manual version number changes still have to be done to both the C code and the Java code.

Fixes #97

Note this also corrects the version to the last release, so when it is bumped again it should update.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -12,6 +12,13 @@ namespace :version do
       "STRSCAN_VERSION \"#{version.succ}\""
     end
     File.write(strscan_c_path, strscan_c)
+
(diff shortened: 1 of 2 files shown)


kou:
Thanks.

PR #99, opened 2024-05-30
Split method descriptions with multibyte strings

nobu:
C compilers which do not default to UTF-8 fail or warn UTF-8 strings.


hsbt:
@kou How about this? This change need to update with https://github.com/ruby/ruby/blob/master/tool/sync_default_gems.rb too. I will update both when this change is merged.

kou:
Sorry. I missed this. I've merged this.

hsbt:
Thanks always 👍

hsbt:
FYI: @nobu said old VC couldn't handle UTF-8 source code because we can't use multi-byte char. But VC on VS 2015 or later can handle that: https://learn.microsoft.com/ja-jp/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170

We may remove this restriction from `ruby/ruby`.

PR #100, opened 2024-05-30
Run test with installed gems

nobu:
As of test-unit-ruby-core 1.0.5, test-unit is not in the dependency list.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
         if: >-
           startsWith(github.ref, 'refs/tags/') &&
             matrix.os == 'ubuntu-latest' &&
-            (matrix.ruby == '3.3' || matrix.ruby == 'jruby')
+            (matrix.ruby == needs.ruby-version.outputs.latest || matrix.ruby == 'jruby')
         run: |
           ruby \
             -e 'print("## strscan "); \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -83,8 +83,6 @@ jobs:
 
       - run: gem install --verbose --backtrace pkg/*.gem
 
-      - run: gem install test-unit-ruby-core
-
       - name: Run tests on the installed gem
-        run: ruby run-test.rb
+        run: bundle exec ruby run-test.rb
```
kou, line 87: If we use `bundle exec` here, we use strscan in the current directory not installed gem:

https://github.com/ruby/strscan/actions/runs/9298585044/job/25590771816?pr=100#step:11:5

```text
Loaded strscan from /home/runner/work/strscan/strscan/lib/strscan.so
```

We want to use strscan installed as gem. For example: We want to use https://github.com/ruby/strscan/actions/runs/9201622389/job/25310137052#step:12:5

```text
Loaded strscan from /opt/hostedtoolcache/Ruby/3.3.1/x64/lib/ruby/gems/3.3.0/gems/strscan-3.1.1/lib/strscan.so
```

nobu:
Add the dependency to test-unit-ruby-core.
https://github.com/ruby/test-unit-ruby-core/pull/5

I expect this failure will be solved by the new gem.

PR #101, opened 2024-05-30
Use C90 as far as supporting 2.6 or earlier

nobu:


ext/strscan/strscan.c
```diff
@@ -795,8 +796,7 @@ strscan_do_scan(VALUE self, VALUE pattern, int succptr, int getstr, int headonly
     }
 
     if (RB_TYPE_P(pattern, T_REGEXP)) {
-        p->regex = pattern;
-        OnigPosition ret = rb_reg_onig_match(pattern,
+        OnigPosition ret = rb_reg_onig_match(p->regex = pattern,
```
kou, line 799: !?

```suggestion
        OnigPosition ret;
        p->regex = pattern;
        ret = rb_reg_onig_match(p->regex,
```

PR #112, opened 2024-10-28
[DOC] doc/strscan/strscan.md - fix triple backtick

MSP-Greg:
See current doc, search for '## Target Substring', notice the backtick that isn't visible to the GitHub md parser.

https://github.com/ruby/strscan/blob/843e931d134b0a8f0284296250454374e3f8a6aa/doc/strscan/strscan.md

Fixed in the PR, see:

https://github.com/MSP-Greg/strscan/blob/00-strscan.md/doc/strscan/strscan.md


kou:
Good catch!

Issue #113, opened 2024-11-07
Provide an efficient way to parse Integers (and Floats)?


casperisfine:
Previous discussion: https://bugs.ruby-lang.org/issues/20394

### Context

When trying to write pure Ruby gems that are competitive in term of performance with C extensions, a very common bottleneck is parsing of text based protocols and formats, such as the [Redis RESP protocol](https://github.com/antirez/RESP3/blob/master/spec.md), or even [the PDF format](https://github.com/gettalong/hexapdf/blob/master/lib/hexapdf/tokenizer.rb#L287-L288) (FYI @gettalong).

As a result, currently [the most efficient way to parse integers in a string in Ruby, is to reimplement `atoi` using `String#getbyte`](https://github.com/redis-rb/redis-client/commit/41b3abe94243d2598211d448c4e457a3585ff9d5), which is a bit ridiculous.

Otherwise if you create a substring with `String#slice` or `StringScanner#scan` and then call `to_i` or `Integer`, instantiating the sub string and copying the bytes really tank the performance.

### Proposal

Given that `StringScanner` is a default gem, is often involved in string parsing, and already act as a "pointer into a String", I think it's well positioned to offer an efficient way to parse an Integer without instantiating a useless temporary string.

Basically an optimized way to do `scanner.scan(/\d+/).to_i`.

The API could be any of:

  - `scanner.scan(/\d+/, :to_i)`
  - `scanner.scan(/\d+/, Integer)`
  - `scanner.scan_integer(/\d+/)`

Logically the two supported types would be `Integer` and `Float`, but perhaps others would be helpful for other protocols?

@kou as maintainer of `strscan`, do you have any opinion? I'm happy to put the work on this, but I'd need to know if the feature is desired, and which API would be deemed acceptable.

Also cc @tenderlove @mame from previous discussions.

tenderlove:
I think it makes sense, but we shouldn't we support things besides base 10?  I'd expect to be able to provide a base (I think)

casperisfine:
The way I was seeing it, these would be a different format name, e.g. with the first syntax:

```
scanner.scan(/\d+/, :to_i)
scanner.scan(/\d[a-f]+/, :hexa)
# ...
```

But maybe that's not flexible enough? The issue is that for this to perform, we need to invoke a C function with a pointer, so argument passing a bit more complex I think?

gettalong:
[...]

PR #114, opened 2024-11-14
Implement #scan_integer to efficiently parse Integer

casperisfine:
Fix: https://github.com/ruby/strscan/issues/113

This allows to directly parse an Integer from a String without needing to first allocate a sub string.

Notes:

The implementation is limited by design, it's meant as a first step, only the most straightforward, based 10 integers are supported.

cc @kou

### DIFF
--- a/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
+++ b/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
@@ -54,6 +54,7 @@ import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
 import org.jruby.util.ByteList;
 import org.jruby.util.StringSupport;
+import org.jruby.util.ConvertBytes;
 
 import java.util.Iterator;
 
@@ -554,6 +555,47 @@ public class RubyStringScanner extends RubyObject {
             runtime.getWarnings().warning(ID.DEPRECATED_METHOD, "StringScanner#peep is obsolete; use #peek instead");
         }
         return peek(context, length);
+    }
+
+    @JRubyMethod(name = "scan_integer")
+    public IRubyObject scan_integer(ThreadContext context) {
+        final Ruby runtime = context.runtime;
+        check(context);
+        clearMatched();
+
+        if (!str.getEncoding().isAsciiCompatible()) {
+            throw getRuntime().newEncodingCompatibilityError("ASCII incompatible encoding: " + str.getEncoding());
+        }
+
+
+        ByteList bytes = str.getByteList();
+        int curr = this.curr;
+
+        int bite = bytes.get(curr);
+        if (bite == '-' || bite == '+') {
+            curr++;
+            bite = bytes.get(curr);
+        }
+
+        if (!(bite >= '0' && bite <= '9')) {
+            return runtime.getNil();
+        }
+
+        while (bite >= '0' && bite <= '9') {
+            curr++;
+            if (curr >= bytes.getRealSize()) {
+                break;
+            }
+            bite = bytes.get(curr);
+        }
+
+        int length = curr - this.curr;
+        prev = this.curr;
+        this.curr = curr;
+        setMatched();
+        adjustRegisters();
+
+        return ConvertBytes.byteListToInum(runtime, bytes, prev, curr, 10, true);
     }
 
     @JRubyMethod(name = "unscan")
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -20,6 +20,7 @@
 extern size_t onig_region_memsize(const struct re_registers *regs);
 #endif
 
+#include <ctype.h>
 #include <stdbool.h>
 
 #define STRSCAN_VERSION "3.1.1"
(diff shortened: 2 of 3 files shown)


casperisfine:
Arf, forgot this gem had a Java version. Looking at it.

casperisfine:
Another thing that in unclear is whether we assume ASCII compatible encoding or not.

casperisfine:
> Another thing that in unclear is whether we assume ASCII compatible encoding or not.

```ruby
>> "1234".encode(Encoding::UTF_32BE).to_i
(irb):2:in `to_i': ASCII incompatible encoding: UTF-32BE (Encoding::CompatibilityError)
```

So I guess we should just do the same here.

casperisfine:
PR updated with the Java implementation, and now raising when called with encoding that's not ASCII compatible.

casperisfine:
Apparently, the Truffle implement is in Truffle itself, so I just skip those test like many others.

kou:
TruffleRuby support isn't included in this repository. We can just ignore TruffleRuby.
See also: #35

ext/strscan/strscan.c
```diff
@@ -1266,6 +1268,55 @@ strscan_peep(VALUE self, VALUE vlen)
...
+    }
+
+    if (!isdigit(ptr[len])) {
+        return Qnil;
+    }
+
+    MATCHED(p);
+    p->prev = p->curr;
+
+    while(isdigit(ptr[len])) {
```
kou, line 1306:
```suggestion
    while (isdigit(ptr[len])) {
```

ext/strscan/strscan.c
```diff
@@ -1266,6 +1268,55 @@ strscan_peep(VALUE self, VALUE vlen)
...
+    struct strscanner *p;
+
+    GET_SCANNER(self, p);
+    CLEAR_MATCH_STATUS(p);
+
+    rb_must_asciicompat(p->str);
+
+    ptr = CURPTR(p);
+
+    if (ptr[len] == '-' || ptr[len] == '+') {
```
kou, line 1295: Is `ptr[len]` safe when `S_RESTLEN()` is `0`?

ext/strscan/strscan.c
```diff
@@ -1266,6 +1268,55 @@ strscan_peep(VALUE self, VALUE vlen)
...
+
+    while(isdigit(ptr[len])) {
+        len++;
+    }
+
+    buffer = ALLOCV_N(char, buffer_v, len + 1);
+
+    MEMCPY(buffer, CURPTR(p), char, len);
+    buffer[len] = '\0';
+    integer = rb_cstr2inum(buffer, 10);
```
kou, line 1314: Ah, we want to avoid this copy...

ext/strscan/strscan.c
```diff
@@ -21,6 +21,7 @@ extern size_t onig_region_memsize(const struct re_registers *regs);
 #endif
 
 #include <stdbool.h>
+#include <ctype.h>
```
kou, line 24: Could you sort this list in alphabetical order?
```suggestion
#include <ctype.h>
#include <stdbool.h>
```

ext/strscan/strscan.c
casperisfine: I think so, because (for now?) Ruby strings are guaranteed to be NULL terminated. So if we're EOF, `ptr[len]` is `\0`.

ext/strscan/strscan.c
kou: OK. We can use `rb_cstr2inum()` for now. Could you propose the fast path to Ruby itself? It may be useful for extensions.

ext/strscan/strscan.c
kou: It seems that a `String` created by `rb_str_new_static()` may not be terminated by `\0`.

ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
```diff
@@ -556,6 +557,47 @@ public IRubyObject peep(ThreadContext context, IRubyObject length) {
...
+
+        if (!str.getEncoding().isAsciiCompatible()) {
+            throw getRuntime().newEncodingCompatibilityError("ASCII incompatible encoding: " + str.getEncoding());
+        }
+
+
+        ByteList bytes = str.getByteList();
+        int curr = this.curr;
+
+        int bite = bytes.get(curr);
```
kou, line 574: `byte`?

olleolleolle: Thanks for asking that, I now re-learned that in Java, there are certain reserved words: https://en.wikipedia.org/wiki/List_of_Java_keywords and `byte` is one of them.

kou: !!!

ext/strscan/strscan.c
byroot: Technically you can but the documentation clearly states:

> Identical to rb_str_new(), except it takes a C string literal.

So it assume it's a literal, hence is NUL terminated.

I can add the checks if you feel strongly about it, but it's a lot of busy code for something I think isn't supposed to happen.

byroot:
Closing in favor of https://github.com/ruby/strscan/pull/115 for logistical reasons.

PR #115, opened 2024-11-25
Implement #scan_integer to efficiently parse Integer

byroot:
Fix: https://github.com/ruby/strscan/issues/113

This allows to directly parse an Integer from a String without needing to first allocate a sub string.

Notes:

The implementation is limited by design, it's meant as a first step, only the most straightforward, based 10 integers are supported.

Reopening https://github.com/ruby/strscan/pull/114 for logistical reasons.


ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
```diff
@@ -556,6 +557,47 @@ public IRubyObject peep(ThreadContext context, IRubyObject length) {
...
     }
 
+    @JRubyMethod(name = "scan_integer")
+    public IRubyObject scan_integer(ThreadContext context) {
+        final Ruby runtime = context.runtime;
+        check(context);
+        clearMatched();
+
+        if (!str.getEncoding().isAsciiCompatible()) {
+            throw getRuntime().newEncodingCompatibilityError("ASCII incompatible encoding: " + str.getEncoding());
```
headius, line 567: runtime variable above can be used here

ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
```diff
@@ -556,6 +557,47 @@ public IRubyObject peep(ThreadContext context, IRubyObject length) {
...
+        int curr = this.curr;
+
+        int bite = bytes.get(curr);
+        if (bite == '-' || bite == '+') {
+            curr++;
+            bite = bytes.get(curr);
+        }
+
+        if (!(bite >= '0' && bite <= '9')) {
+            return runtime.getNil();
```
headius, line 581: context.nil is a quicker path to the nil object

headius (approved):
Code looks great, thanks for adding it! Only two minor "tips" in my review.

ext/strscan/strscan.c
```diff
@@ -1266,6 +1268,61 @@ strscan_peep(VALUE self, VALUE vlen)
...
+    while (len < remaining_len && isdigit(ptr[len])) {
+        len++;
+    }
+
+    buffer = ALLOCV_N(char, buffer_v, len + 1);
+
+    MEMCPY(buffer, CURPTR(p), char, len);
+    buffer[len] = '\0';
+    integer = rb_cstr2inum(buffer, 10);
+    RB_GC_GUARD(buffer_v);
```
kou, line 1321: Why do we need to call `RB_GC_GUARD()` here?
Did you want to use `RB_ALLOCV_END()` instead?

ext/strscan/strscan.c
```diff
@@ -1266,6 +1268,61 @@ strscan_peep(VALUE self, VALUE vlen)
...
+    }
+
+    MATCHED(p);
+    p->prev = p->curr;
+
+    while (len < remaining_len && isdigit(ptr[len])) {
+        len++;
+    }
+
+    buffer = ALLOCV_N(char, buffer_v, len + 1);
```
kou, line 1316: Can we use `RB_ALLOCV_N()` instead?

byroot: Done

kou:
Thanks.

byroot:
Thank you for the merge.

Let me know what you desire as next steps. I think we somewhat identified the need for a `base:` argument, and perhaps a `pattern:` one?

I suppose they are kinda tied together, e.g. to parse things like `0x123DEF`, you may call `scan_integer(pattern: /0x[0-9a-f]+/, base: 16)` ?

kou:
How about `scan_integer(base: 16)` that accepts `/0x[0-9a-fA-F]+/` as the next step?
What patterns should we accept by default for `base: 16`?
Do we have use cases for other patterns for now?

byroot:
> What patterns should we accept by default for base: 16?

I would say `(0x)?[0-9a-fA-F]+` ?

> Do we have use cases for other patterns for now?

Personally I don't. 

One question though, if we accept `base: 16`, should we error on anything other than `base: 10` and `base: 16`?

kou:
> > What patterns should we accept by default for base: 16?
> 
> I would say `(0x)?[0-9a-fA-F]+` ?

OK. Let's use it.
 
> One question though, if we accept `base: 16`, should we error on anything other than `base: 10` and `base: 16`?

Yes. We may add `base: 8` or something but we should raise an error for now.

PR #116, opened 2024-11-26
StringScanner#scan_integer support base 16 integers

byroot:
Followup: https://github.com/ruby/strscan/pull/115

`scan_integer` is now implemented in Ruby as to efficiently handle keyword arguments without allocating a Hash. Given the goal of `scan_integer` is to more effciently parse integers without having to allocate an intermediary object, using `rb_scan_args` would defeat the purpose.

Additionally, the C implementation now uses `rb_isdigit` and `rb_isxdigit`, because on Windows `isdigit` is locale dependent.

cc @kou

### DIFF
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -20,7 +20,6 @@
 extern size_t onig_region_memsize(const struct re_registers *regs);
 #endif
 
-#include <ctype.h>
 #include <stdbool.h>
 
 #define STRSCAN_VERSION "3.1.1"
@@ -116,7 +115,7 @@ static VALUE strscan_get_byte _((VALUE self));
 static VALUE strscan_getbyte _((VALUE self));
 static VALUE strscan_peek _((VALUE self, VALUE len));
 static VALUE strscan_peep _((VALUE self, VALUE len));
-static VALUE strscan_scan_integer _((VALUE self));
+static VALUE strscan_scan_base10_integer _((VALUE self));
 static VALUE strscan_unscan _((VALUE self));
 static VALUE strscan_bol_p _((VALUE self));
 static VALUE strscan_eos_p _((VALUE self));
@@ -1268,21 +1267,26 @@ strscan_peep(VALUE self, VALUE vlen)
     return strscan_peek(self, vlen);
 }
 
-/*
- * call-seq:
- *   scan_integer
- *
- * Equivalent to #scan with a [+-]?\d+ pattern, and returns an Integer or nil.
- *
- * The scanned string must be encoded with an ASCII compatible encoding, otherwise
- * Encoding::CompatibilityError will be raised.
- */
 static VALUE
-strscan_scan_integer(VALUE self)
+strscan_parse_integer(struct strscanner *p, int base, long len)
 {
-    char *ptr, *buffer;
-    long len = 0;
     VALUE buffer_v, integer;
+
+    char *buffer = RB_ALLOCV_N(char, buffer_v, len + 1);
+
+    MEMCPY(buffer, CURPTR(p), char, len);
+    buffer[len] = '\0';
+    integer = rb_cstr2inum(buffer, base);
+    RB_ALLOCV_END(buffer_v);
+    p->curr += len;
+    return integer;
+}
+
+static VALUE
+strscan_scan_base10_integer(VALUE self)
+{
+    char *ptr;
+    long len = 0;
     struct strscanner *p;
 
     GET_SCANNER(self, p);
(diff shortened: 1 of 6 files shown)


lib/strscan.rb
```diff
@@ -0,0 +1,32 @@
...
+end
+
+class StringScanner
+  # call-seq:
+  #   scan_integer(base: 10)
+  # 
+  # If `base` isn't provided or is `10`, then it is equivalent to calling `#scan` with a `[+-]?\d+` pattern,
+  # and returns an Integer or nil.
+  #
+  # If `base` is `16`, then it is equivalent to calling `#scan` with a `(0x)?[0-9a-fA-F]+` pattern,
```
kou, line 17: Could you accept sign?
`[+-](0x)?[0-9a-fA-F]+`

It seems sign can be used: https://en.wikipedia.org/wiki/Hexadecimal#Signs

strscan.gemspec
```diff
@@ -19,14 +19,14 @@ Gem::Specification.new do |s|
   files = [
     "COPYING",
     "LICENSE.txt",
-  ]
+  ] + Dir[File.join(__dir__, "lib/**/*.rb")]
```
kou, line 22: Could you use a static list? Glob may not work in ruby/ruby.

```suggestion
    "lib/strscan.rb",
  ]
```

byroot: Done.

lib/strscan.rb
byroot: Done.

byroot:
Hum, Truffle CI is broken, because now `lib/strscan.rb` is shadowing the `lib/truffle/strscan.rb` that's embeded by Truffle.

@eregon any idea how to fix it? Is there even value in testing TruffleRuby in this repo if the implementation is in Truffle itself?

byroot:
I pushed https://github.com/ruby/strscan/pull/116/commits/caaaf94c04026bdc6c0621eba0a8dddf92b7fd96 to fix TruffleRuby CI, but it's quite ugly.

kou:
I want to remove TruffleRuby related codes from this repository but how about moving `lib/strscan.rb` to `lib/strscan/strscan.rb` and calling `require "strscan/strscan.rb"` in `Init_strscan()`?

byroot:
> how about moving `lib/strscan.rb` to `lib/strscan/strscan.rb` and calling `require "strscan/strscan.rb"` in `Init_strscan()`?

If that's what you prefer, it's possible yes.

ext/strscan/strscan.c
```diff
@@ -2290,4 +2330,6 @@ Init_strscan(void)
     rb_define_method(StringScanner, "fixed_anchor?", strscan_fixed_anchor_p, 0);
 
     rb_define_method(StringScanner, "named_captures", strscan_named_captures, 0);
+
+    rb_require("strscan/strscan");
```
kou, line 2334: We may need to use `rb_funcall(rb_mKernel, rb_intern("require"), 1, rb_str_new_cstr("strscan/strscan"))` or something because RubyGems may replace `rb_require()`.

byroot: Are you sure? Other gems like `psych`, `json` etc use `rb_require` directly without known issues.

kou: Oh, I didn't know it.
OK. Let's use this. I hope that this is a temporary workaround...

kou:
Thanks.

byroot:
Thank you. Let me know if there any more features you'd like for the initial version, and if some use cases come up in the future I'm happy to help implement them.

kou:
I don't have any more requests for the initial version.

I'll release a new version in a few weeks.

byroot:
For the record:

```ruby
# frozen_string_literal: true
require 'strscan'
require 'benchmark/ips'

source = 10_000.times.map { rand(9999999).to_s }.join(",").freeze


def scan_to_i(source)
  scanner = StringScanner.new(source)
  while number = scanner.scan(/\d+/)
    number.to_i
    scanner.skip(",")
  end
end

def scan_integer(source)
  scanner = StringScanner.new(source)
  while scanner.scan_integer
    scanner.skip(",")
  end
end


Benchmark.ips do |x|
  x.report("scan.to_i") { scan_to_i(source) }
  x.report("scan_integer") { scan_integer(source) }
  x.compare!
end
```

```
$ bundle exec ruby --yjit /tmp/bench-scan.rb 
ruby 3.3.4 (2024-07-09 revision be1089c8ec) +YJIT [arm64-darwin23]
           scan.to_i    90.000 i/100ms
        scan_integer   215.000 i/100ms
           scan.to_i    907.964 (± 1.1%) i/s    (1.10 ms/i) -      4.590k in   5.055944s
        scan_integer      2.144k (± 0.2%) i/s  (466.37 μs/i) -     10.750k in   5.013472s

Comparison:
        scan_integer:     2144.2 i/s
           scan.to_i:      908.0 i/s - 2.36x  slower
```

I was hopping for more, but 2.3x is already a nice gain, and profile shows there is some optimizations we could do in the Ruby side: https://share.firefox.dev/49bOcTn

PR #126, opened 2024-12-12
[DOC] Add syntax highlighting to MarkDown code blocks

amomchilov:
Split off from https://github.com/ruby/ruby/pull/12322


kou:
Thanks.

Issue #125, opened 2024-12-12
JRuby: test_scan_integer_base_16 failed only with ubuntu-latest + jruby-head


kou:
https://github.com/ruby/strscan/actions/runs/12288841240/job/34293324916#step:11:10

```text
Error: test_scan_integer_base_16(TestStringScanner): Java::JavaLang::ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
org.jruby.dist/org.jruby.util.ByteList.get(ByteList.java:632)
org.jruby.ext.strscan.RubyStringScanner.scan_base16_integer(RubyStringScanner.java:621)
org.jruby.ext.strscan.RubyStringScanner$INVOKER$i$0$0$scan_base16_integer.call(RubyStringScanner$INVOKER$i$0$0$scan_base16_integer.gen)
org.jruby.dist/org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:456)
org.jruby.dist/org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:195)
org.jruby.dist/org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:346)
org.jruby.dist/org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:66)
org.jruby.dist/org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:82)
org.jruby.dist/org.jruby.internal.runtime.methods.MixedModeIRMethod.INTERPRET_METHOD(MixedModeIRMethod.java:201)
org.jruby.dist/org.jruby.internal.runtime.methods.MixedModeIRMethod.call(MixedModeIRMethod.java:188)
org.jruby.dist/org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:220)
org.jruby.dist/org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:466)
org.jruby.dist/org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:244)
org.jruby.dist/org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:314)
org.jruby.dist/org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:66)
[...]

PR #127, opened 2024-12-12
Re-port new integer-scanning utility methods

headius:
This is a re-port of the C code from @byroot for fast base 10 and base 16 integer scanning.

In https://github.com/ruby/strscan/issues/125, @kou pointed out there's an intermittent failure in the JRuby extension. We were unable to confirm exactly the circumstances that cause that failure, but this re-port should at least help reduce the change it is a bug in the original Java code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,9 @@ on:
 - push
 - pull_request
 
+env:
+  JRUBY_OPTS: "-X+C" # temporarily force JRuby to compile, so Java exception trace will contain .rb lines
+
 jobs:
   ruby-versions:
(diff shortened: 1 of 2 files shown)


headius:
@kou @byroot this should be ready to go but I have one concern about the C code...

It doesn't seem to update `curr` after the parse? I have that in the Java code because it seemed necessary.

byroot:
> It doesn't seem to update `curr` after the parse?

It does inside `strscan_parse_integer`

headius:
> It does inside strscan_parse_integer

Aha, so it does! I will tweak this to move the final integer parse and `curr` update into a similar method, so it aligns better with the C code.

kou:
Thanks!

Issue #130, opened 2025-02-10
#matched after #scan_integer causes segfault


andrykonchin:
I've reproduced the following issue on Ruby 3.4.1, strscan version is the most recent - 3.1.2, macOS:

```ruby
require 'strscan'
s = StringScanner.new("42abc");
s.scan_integer
s.matched
```

### Output

<details>

<summary>Crash Report log information</summary>

```
$ ruby test.rb
test.rb:4: [BUG] Segmentation fault at 0x0000000000000000
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-darwin23]

-- Crash Report log information --------------------------------------------
   See Crash Report log file in one of the following locations:
     * ~/Library/Logs/DiagnosticReports
     * /Library/Logs/DiagnosticReports
   for more details.
Don't forget to include the above Crash Report log file in bug reports.

-- Control frame information -----------------------------------------------
c:0003 p:---- s:0011 e:000010 CFUNC  :matched
c:0002 p:0023 s:0007 E:001308 EVAL   test.rb:4 [FINISH]
c:0001 p:0000 s:0003 E:001290 DUMMY  [FINISH]

-- Ruby level backtrace information ----------------------------------------
test.rb:4:in '<main>'
test.rb:4:in 'matched'

-- Threading information ---------------------------------------------------
Total ractor count: 1
Ruby thread count for this ractor: 1

-- Machine register context ------------------------------------------------
 rax: 0x0000000000000000 rbx: 0x0000600003974000 rcx: 0x0000000109034038
 rdx: 0x0000000109034038 rdi: 0x000000010912d3b0 rsi: 0x0000000109034038
 rbp: 0x00007ff7b6f5b9a0 rsp: 0x00007ff7b6f5b990  r8: 0x00007fd0c231b040
  r9: 0x00007fd0c231b038 r10: 0x33f9482520000031 r11: 0x00006000018702a0
 r12: 0x000000010912d130 r13: 0x0000600001475ad0 r14: 0x000000010912d130
 r15: 0x00007fd0c241af90 rip: 0x0000000109032539 rfl: 0x0000000000010202

-- C level backtrace information -------------------------------------------
/Users/andrykonchin/.rbenv/versions/3.4.1/lib/libruby.3.4.dylib(rb_vm_bugreport+0x9bd) [0x109eb95fd]
[...]

PR #133, opened 2025-02-12
Update registers after parsing integers

byroot:
Fix GH-130

### DIFF
--- a/test/strscan/test_stringscanner.rb
+++ b/test/strscan/test_stringscanner.rb
@@ -946,6 +946,18 @@ module StringScannerTests
     end
   end
 
+  def test_scan_integer_matched
+    omit("not implemented on TruffleRuby") if RUBY_ENGINE == "truffleruby"
+
+    s = create_string_scanner("42abc")
+    assert_equal(42, s.scan_integer)
+    assert_equal("42", s.matched)
+
+    s = create_string_scanner("42abc")
+    assert_equal(0x42abc, s.scan_integer(base: 16))
+    assert_equal("42abc", s.matched)
+  end
+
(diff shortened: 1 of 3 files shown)


byroot:
JRuby needs the same fix it seems.

byroot:
> JRuby needs the same fix it seems.

Done.

test/strscan/test_stringscanner.rb
```diff
@@ -946,6 +946,16 @@ def test_scan_integer_encoding
     end
   end
 
+  def test_scan_integer_matched
```
eregon, line 949: You need `omit("scan_integer isn't implemented on TruffleRuby yet") if RUBY_ENGINE == "truffleruby"` as well here (will address those soon, TruffleRuby is implementing the new method)

test/strscan/test_stringscanner.rb
```diff
@@ -946,6 +946,18 @@ def test_scan_integer_encoding
...
+    omit("not implemented on TruffleRuby") if RUBY_ENGINE == "truffleruby"
+
+    s = create_string_scanner("42abc")
+    assert_equal(42, s.scan_integer)
+    assert_equal("42", s.matched)
+
+    s = create_string_scanner("42abc")
+    assert_equal(273084, s.scan_integer(base: 16))
```
kou, line 957: How about using hex literal for easy to read?

```suggestion
    assert_equal(0x42abc, s.scan_integer(base: 16))
```

kou (approved):
+1

byroot:
Thanks for the review. Updated.

PR #134, opened 2025-02-12
Add rdoc to Gemfile to fix ruby-head builds

byroot:


Gemfile
```diff
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 gemspec
 
 group :development do
+  gem "rdoc"
```
kou, line 6: Could you keep this list in alphabetical order?

byroot: Done.

Issue #135, opened 2025-02-12
Scanning methods that don't use Regexp don't clear named capture groups


andrykonchin:
I noticed strange behaviour - `#named_captures` may return capture groups  and `#[]` returns `nil` instead of raising `IndexError` in the following scenario:
1) call a scanning method with a Regexp (with named capture groups)
2) call a scanning method without a Regexp pattern (`#getch`, `#get_byte` or `#scan_until`/`#exist?`/... with a String pattern)
3) call `#[]` with a named capture group (used in the Regexp) and `#named_captures`

I would expect that `#[]` raises `IndexError` and `#named_captures` returns `nil`.

Example

```ruby
require 'strscan'

s = StringScanner.new('Fri Dec 12 1975 14:39')

s.scan(/(?<wday>Fri)/)
p s.named_captures # => {"wday" => "Fri"}
p s["wday"] # => "Fri"

s.get_byte
p s.named_captures # => {"wday" => nil}
p s["wday"] # => nil

s["month"] # => undefined group name reference: month (IndexError)
```

Wondering whether it's correct behaviour.

kou:
Good catch!

We should clear named captures.

naitoh:
> Do you want to work on this too?

Yes, I would like to try this issue, too.

PR #142, opened 2025-02-22
Fix a bug that scanning methods that don't use Regexp don't clear named capture groups

naitoh:
Fix GH-135

### DIFF
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -58,8 +58,13 @@ struct strscanner
 };
 
 #define MATCHED_P(s)          ((s)->flags & FLAG_MATCHED)
-#define MATCHED(s)             (s)->flags |= FLAG_MATCHED
-#define CLEAR_MATCH_STATUS(s)  (s)->flags &= ~FLAG_MATCHED
+#define MATCHED(s)            ((s)->flags |= FLAG_MATCHED)
+#define CLEAR_MATCHED(s)      ((s)->flags &= ~FLAG_MATCHED)
+#define CLEAR_NAMED_CAPTURES(s) ((s)->regex = Qnil)
+#define CLEAR_MATCH_STATUS(s) do {\
+    CLEAR_MATCHED(s);\
+    CLEAR_NAMED_CAPTURES(s);\
+} while (0)
 
 #define S_PBEG(s)  (RSTRING_PTR((s)->str))
 #define S_LEN(s)  (RSTRING_LEN((s)->str))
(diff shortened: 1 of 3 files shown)


test/strscan/test_stringscanner.rb
```diff
@@ -944,6 +944,60 @@ def test_named_captures
...
     assert_equal({"f" => "foo", "r" => "bar", "z" => "baz"}, scan.named_captures)
+
+    # test that named_captures has been reset
+    assert_equal(9, scan.match?("foobarbaz"))
+    assert_equal({}, scan.named_captures)
+
+    assert_equal(9, scan.match?(/(?<f>foo)(?<r>bar)(?<z>baz)/))
+    assert_equal({"f" => "foo", "r" => "bar", "z" => "baz"}, scan.named_captures)
+    scan.unscan
+    assert_equal({}, scan.named_captures)
```
kou, line 955: Could you move this to `test_unscan`? (Could you move other tests too?)

ext/strscan/strscan.c
```diff
@@ -361,6 +361,7 @@ strscan_reset(VALUE self)
 
     GET_SCANNER(self, p);
     p->curr = 0;
+    p->regex = Qnil;
```
kou, line 364: We need to do this in all locations that use `CLEAR_MATCH_STATUS(p)`, right?
How about adding a new macro or function that reset `p->regex` and call `CLEAR_MATCH_STATUS(p)` and replacing `CLEAR_MATCH_STATUS(p)` calls with it?

test/strscan/test_stringscanner.rb
naitoh: OK.
I see.

ext/strscan/strscan.c
naitoh: Yes.
I will use new macro.

ext/strscan/strscan.c
```diff
@@ -58,8 +58,10 @@ struct strscanner
 };
 
 #define MATCHED_P(s)          ((s)->flags & FLAG_MATCHED)
-#define MATCHED(s)             (s)->flags |= FLAG_MATCHED
-#define CLEAR_MATCH_STATUS(s)  (s)->flags &= ~FLAG_MATCHED
+#define MATCHED(s)            ((s)->flags |= FLAG_MATCHED)
+#define CLEAR_MATCHED(s)      ((s)->flags &= ~FLAG_MATCHED)
+#define CLEAR_NAMED_CAPTURES(s) ((s)->regex = Qnil)
+#define CLEAR_MATCHED_AND_NAMED_CAPTURES(s) CLEAR_MATCHED(s); CLEAR_NAMED_CAPTURES(s)
```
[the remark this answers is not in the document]
naitoh: OK,
I see.

kou:
Thanks.

Issue #139, opened 2025-02-18
Inconsistency of IndexError vs nil for unknown capture group


eregon:
After matching with a String, `StringScanner#[:unknown_capture_group]` returns `nil`:
```
$ irb -rstrscan
irb(main):001> s=StringScanner.new("abc")
=> #<StringScanner 0/3 @ "abc">
irb(main):002> s.scan("a")
=> "a"
irb(main):003> s[:foo]
=> nil
```

After matching with a Regexp, `StringScanner#[:unknown_capture_group]` raises IndexError:
(using a new process to be sure to not be affected by https://github.com/ruby/strscan/issues/135)
```
irb -rstrscan
irb(main):001> s=StringScanner.new("abc")
=> #<StringScanner 0/3 @ "abc">
irb(main):002> s.scan(/./)
=> "a"
irb(main):003> s[:foo]
(irb):3:in `[]': undefined group name reference: foo (IndexError)
```

It would be best if the behavior is consistent.

There is a test for the `nil` case in https://github.com/ruby/strscan/blob/v3.1.2/test/strscan/test_stringscanner.rb#L830-L840
And there are some tests for the IndexError cases in https://github.com/ruby/strscan/blob/v3.1.2/test/strscan/test_stringscanner.rb#L443-L444 and in more places.

The implementation is https://github.com/ruby/strscan/blob/6968b119a684a086d67af079db4d77a05907de85/ext/strscan/strscan.c#L1764

This is proving problematic on TruffleRuby.
We would like to upstream our pure-Ruby implementation of StringScanner, but to do so we want to avoid exposing internals as much as possible.
Things like `p->regex` is something I don't want to expose to code living outside the truffleruby repository.
Having to workaround by keeping extra state for whether the last match was with a String or a Regexp feels very messy and wrong semantically.

Could the behavior be unified for both of these cases?

kou:
OK. Let's use `IndexError` for `String` too.

@naitoh Do you want to work on this?

naitoh:
> Do you want to work on this?

Yes, I would like to try it.

eregon:
I think the logic should be:
* There is always a "result of the last StringScanner match" (some kind of MatchData internally or `nil`) whether matched against String pattern or Regexp pattern
* If the last match failed, the last result is nil, then `[]` returns `nil` for any group
* If the last match succeeded, then use the semantics of `MatchData#[]` or equivalent, which raises `IndexError` on missing group. For a String pattern there are of course no named groups.

[...]

PR #143, opened 2025-02-22
Fix a bug that inconsistency of IndexError vs nil for unknown capture group

naitoh:
Fix GH-139

### DIFF
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -1667,19 +1667,17 @@ strscan_matched_size(VALUE self)
 static int
 name_to_backref_number(struct re_registers *regs, VALUE regexp, const char* name, const char* name_end, rb_encoding *enc)
 {
-    int num;
-
-    num = onig_name_to_backref_number(RREGEXP_PTR(regexp),
-	(const unsigned char* )name, (const unsigned char* )name_end, regs);
-    if (num >= 1) {
-	return num;
+    if (RTEST(regexp)) {
+        int num = onig_name_to_backref_number(RREGEXP_PTR(regexp),
+                                              (const unsigned char* )name,
+                                              (const unsigned char* )name_end,
+                                              regs);
+        if (num >= 1) {
+	        return num;
+        }
     }
-    else {
-	rb_enc_raise(enc, rb_eIndexError, "undefined group name reference: %.*s",
-					  rb_long2int(name_end - name), name);
-    }
-
-    UNREACHABLE;
+    rb_enc_raise(enc, rb_eIndexError, "undefined group name reference: %.*s",
+                 rb_long2int(name_end - name), name);
 }
 
 /*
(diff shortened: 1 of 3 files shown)


ext/strscan/strscan.c
```diff
@@ -1663,6 +1663,9 @@ strscan_matched_size(VALUE self)
 static int
 name_to_backref_number(struct re_registers *regs, VALUE regexp, const char* name, const char* name_end, rb_encoding *enc)
 {
+    if (!RTEST(regexp))
+        rb_enc_raise(enc, rb_eIndexError, "undefined group name reference: %.*s",
+                     rb_long2int(name_end - name), name);
     int num;
 
     num = onig_name_to_backref_number(RREGEXP_PTR(regexp),
```
kou, line 1671: Can we reuse existing `rb_enc_raise()`?

```suggestion
    int num = -1;

    if (RTEST(regexp) {
      num = onig_name_to_backref_number(RREGEXP_PTR(regexp),
```

kou (commented):
Do we have `scanner.scan("not-matched")` case?

ext/strscan/strscan.c
naitoh: It's nice!

naitoh:
> Do we have `scanner.scan("not-matched")` case?

Added.

In the `not-matched` case, `nil` return without exception.
But since the results of `Regexp pattern` and `String pattern` are the same, so I think no problem.

ext/strscan/strscan.c
```diff
@@ -1669,17 +1669,15 @@ name_to_backref_number(struct re_registers *regs, VALUE regexp, const char* name
...
 
-    num = onig_name_to_backref_number(RREGEXP_PTR(regexp),
-	(const unsigned char* )name, (const unsigned char* )name_end, regs);
-    if (num >= 1) {
-	return num;
-    }
-    else {
-	rb_enc_raise(enc, rb_eIndexError, "undefined group name reference: %.*s",
-					  rb_long2int(name_end - name), name);
+    if (RTEST(regexp)){
```
kou, line 1672:
```suggestion
    if (RTEST(regexp)) {
```

ext/strscan/strscan.c
```diff
@@ -1669,17 +1669,15 @@ name_to_backref_number(struct re_registers *regs, VALUE regexp, const char* name
...
+        num = onig_name_to_backref_number(RREGEXP_PTR(regexp),
```
kou, line 1673: Can we narrow `num` scope?

```suggestion
        int num = onig_name_to_backref_number(RREGEXP_PTR(regexp),
```

ext/strscan/strscan.c
```diff
@@ -1669,17 +1669,15 @@ name_to_backref_number(struct re_registers *regs, VALUE regexp, const char* name
...
+	        (const unsigned char* )name, (const unsigned char* )name_end, regs);
+        if (num >= 1) {
+	        return num;
+        }
     }
-
-    UNREACHABLE;
+    rb_enc_raise(enc, rb_eIndexError, "undefined group name reference: %.*s",
+        rb_long2int(name_end - name), name);
```
kou, line 1680:
```suggestion
                 rb_long2int(name_end - name), name);
```

ext/strscan/strscan.c
naitoh: done

ext/strscan/strscan.c
naitoh: done

ext/strscan/strscan.c
naitoh: done

kou:
Thanks.

PR #144, opened 2025-02-24
Enable tests passing on TruffleRuby

andrykonchin:
Changes:
- enabled tests passing on TruffleRuby
- removed `truffleruby` and keep only `truffleruby-head` in CI


eregon (approved):
LGTM

eregon:
Context: we recently implemented all the new StringScanner methods in TruffleRuby.
Next we plan to upstream https://github.com/oracle/truffleruby/blob/master/lib/truffle/strscan.rb in this repository after improving it to avoid depending on internals.

kou:
Great!

Issue #145, opened 2025-03-24
#named_captures with regexp union does not match regexp.match behavior


yxhuvud:
Given the following:
```
require "strscan"
scanner = StringScanner.new('foobar')

re1 = /(?<foo>foo)/
re2 = /(?<foo>bar)/

re = Regexp.union([re1, re2])
m = scanner.scan(re)
p m
p scanner.captures
p scanner.named_captures
p 'foobar'.match(re).named_captures
```

It will output: 
```
foo"
["foo", nil]
{"foo" => nil}
{"foo" => "foo"}
```

Removing `re2` from the union will have the scanner named captures as `{"foo" => "foo"}` as expected.

edit: Using `re = /#{re1}|#{re2}/` also seems to have this behavior.

eregon:
As a note, TruffleRuby already has the correct behavior here.
Also, https://bugs.ruby-lang.org/issues/21288 was filed about this too.

PR #146, opened 2025-03-25
named_captures: fix incompatibility with MatchData#named_captures

kou:
Fix GH-145

`MatchData#named_captures` use the last matched value for each name.

Reported by Linus Sellberg. Thanks!!!


byroot:
Also fixes: https://bugs.ruby-lang.org/issues/21288

kou:
Oh, I forgot to merge this...

PR #147, opened 2025-04-10
Allow parsing strings larger than 2GiB

casperisfine:
For a reason unknown, even though `pos` is stored as a `long`, the `#pos` and `#pos=` treat it as an `int`, which prevent seeking into strings larger than 2GiB.

cc @kou

### DIFF
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -524,7 +524,7 @@ strscan_get_pos(VALUE self)
     struct strscanner *p;
 
     GET_SCANNER(self, p);
-    return INT2FIX(p->curr);
+    return LONG2NUM(p->curr);
 }
 
 /*
@@ -554,7 +554,7 @@ strscan_set_pos(VALUE self, VALUE v)
     long i;
 
     GET_SCANNER(self, p);
-    i = NUM2INT(v);
+    i = NUM2LONG(v);
     if (i < 0) i += S_LEN(p);
     if (i < 0) rb_raise(rb_eRangeError, "index out of range");
     if (i > S_LEN(p)) rb_raise(rb_eRangeError, "index out of range");


ext/strscan/strscan.c
```diff
@@ -524,7 +524,7 @@ strscan_get_pos(VALUE self)
     struct strscanner *p;
 
     GET_SCANNER(self, p);
-    return INT2FIX(p->curr);
+    return LONG2FIX(p->curr);
```
kou, line 527: Do we need to use `LONG2NUM()` not `LONG2FIX()`...?

```suggestion
    return LONG2NUM(p->curr);
```

test/strscan/test_stringscanner.rb
```diff
@@ -1128,6 +1128,15 @@ def test_fixed_anchor_false
     assert_equal(false, StringScanner.new("a", fixed_anchor: nil).fixed_anchor?)
     assert_equal(false, StringScanner.new("a", fixed_anchor: false).fixed_anchor?)
   end
+
+  def test_large_string
+    six_gib = (6 * 1024 * 1024 * 1024)
+    string = "a" * six_gib
+    s = create_string_scanner(string) # 6 GiB string
```
kou, line 1135: Does this mean that we need at least 6 GiB memory to run this test...?

ext/strscan/strscan.c
casperisfine: Corrected thanks.

test/strscan/test_stringscanner.rb
casperisfine: Indeed. I don't know how to test it otherwise. I could reduce the size to a bit more than 4GiB, or even 2GiB given it's signed.

We could also just not test it, but this revealed that the Java version has the same limitation, so I think it was valuable. (I won't update the JRuby version though, because there everything uses an int so it's way to much work for me. cc @headius)

casperisfine: I ended up removing the test, as it also doesn't work on windows, nor on 32bit systems.

kou: OK.

PR #150, opened 2025-05-03
Pass end index to byteListToInum

headius:
These parse methods take begin and end indices, not begin and length. A test is included.

Fixes https://github.com/jruby/jruby/issues/8823

### DIFF
--- a/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
+++ b/ext/jruby/org/jruby/ext/strscan/RubyStringScanner.java
@@ -651,7 +651,7 @@ public class RubyStringScanner extends RubyObject {
         setMatched();
         adjustRegisters();
 
-        return ConvertBytes.byteListToInum(runtime, bytes, ptr, len, base, true);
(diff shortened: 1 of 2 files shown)


kou:
Should we release a new version soon?

headius:
@kou If it's convenient, that would be great! I don't have any other changes coming right now.

Also I think this fixes the matrix for windows-latest and "jruby", because the released "jruby" now is JRuby 10 and requires Java 21+:

``diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 7890e97..6034394 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
           distribution: zulu
           java-version: 21
         if: >-
-          matrix.ruby == 'jruby-head'
+          startWith(matrix.ruby, 'jruby')
 
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
```

kou:
Could you open a new PR for the CI change?
I'll release a new version after we merge the CI change.

PR #151, opened 2025-05-03
Fix JRuby CI for Windows

headius:
The setup-ruby logic for JRuby on Windows must not be fully choosing Java 21, so when it tries to compile the extension it's still using an older version. This fixes it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
           distribution: zulu
           java-version: 21
         if: >-
-          matrix.ruby == 'jruby-head'
+          startsWith(matrix.ruby, 'jruby')
 
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1


headius:
I mean I'll get it running and then let you merge this. 😀

headius:
@kou Please hold off on release... I may have found one more bug running HexaPDF tests.

eregon:
> The setup-ruby logic for JRuby on Windows must not be fully choosing Java 21, so when it tries to compile the extension it's still using an older version. This fixes it.

@headius Could you make a PR to setup-ruby to fix that? It would be good to avoid adding many `actions/setup-java` everywhere just to workaround this bug.

eregon:
Mmh, it should work fine since https://github.com/ruby/setup-ruby/blob/eaecf785f6a34567a6d97f686bbb7bccc1ac1e5c/ruby-builder.js#L59 is done on Windows too.
And this is the failing job: https://github.com/ruby/strscan/actions/runs/14805467215/job/41572923725
But as we see on [this line](https://github.com/ruby/strscan/actions/runs/14805467215/job/41572923725#step:5:10) JAVA_HOME is set correctly.

I suppose `rake compile` for Java extensions might need `$JAVA_HOME/bin` to be in PATH, and not just having JAVA_HOME set then (makes sense if e.g. it calls out to `javac`).
So I guess then we should add `$JAVA_HOME/bin` in PATH too in setup-ruby for JRuby (a bit unfortunate as it might cause unintended side effects but I don't see another solution), PR welcome.

kou:
Ah, `RbConfig::CONFIG["EXEEXT"]` may be related. Could you try rake-compiler 1.3.1?

eregon:
https://github.com/ruby/strscan/pull/155

Issue #152, opened 2025-05-03
jruby: ArrayIndexOutOfBounds running HexaPDF tokenizer tests


headius:
Something still not quite right in the `scan_integer` logic.

Here's the full trace:

```
HexaPDF::Content::Tokenizer#test_0007_next_token: returns a PDF keyword for a solitary plus sign:
Java::JavaLang::ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
    org.jruby.dist/org.jruby.util.ByteList.get(ByteList.java:632)
    org.jruby.ext.strscan.RubyStringScanner.scan_base10_integer(RubyStringScanner.java:592)
    Users.headius.work.jruby.lib.ruby.gems.shared.gems.strscan_minus_3_dot_1_dot_5_minus_java.lib.strscan.strscan.️❤ def scan_integer #1(/Users/headius/work/jruby/lib/ruby/gems/shared/gems/strscan-3.1.5-java/lib/strscan/strscan.rb:18)
    Users.headius.work.hexapdf.lib.hexapdf.content.parser.️❤ def parse_number #8(lib/hexapdf/content/parser.rb:135)
    Users.headius.work.hexapdf.lib.hexapdf.content.parser.️❤ def parse_number #8 **(lib/hexapdf/content/parser.rb:131)
    org.jruby.dist/org.jruby.internal.runtime.methods.CompiledIRMethod.call(CompiledIRMethod.java:139)
    org.jruby.dist/org.jruby.ir.targets.indy.InvokeSite.performIndirectCall(InvokeSite.java:860)
    org.jruby.dist/org.jruby.ir.targets.indy.InvokeSite.invoke(InvokeSite.java:808)
    Users.headius.work.hexapdf.lib.hexapdf.content.parser.️❤ def next_token #7(lib/hexapdf/content/parser.rb:90)
    Users.headius.work.hexapdf.lib.hexapdf.content.parser.️❤ def next_token #7 **(lib/hexapdf/content/parser.rb:85)
    org.jruby.dist/org.jruby.internal.runtime.methods.CompiledIRMethod.call(CompiledIRMethod.java:139)
    org.jruby.dist/org.jruby.ir.targets.indy.InvokeSite.performIndirectCall(InvokeSite.java:860)
    org.jruby.dist/org.jruby.ir.targets.indy.InvokeSite.invoke(InvokeSite.java:785)
[...]

PR #153, opened 2025-05-03
jruby: Check if len++ walked off the end

headius:
Fix GH-152

CRuby can walk off the end because there's always a null byte. In JRuby, the byte array is often (usually?) the exact size of the string. So we need to check if len++ walked off the end.

This code was ported from a version by @byroot in https://github.com/ruby/strscan/pull/127 but I missed adding this check due to a lack of tests. A test is included for both "-" and "+" parsing.


headius:
cc @gettalong @kou I think this is the last strscan-related fix from HexaPDF tests.

headius:
JRuby CI failures will be fixed by #151.

kou:
Could you rebase on main?

kou:
Released.

PR #156, opened 2025-05-24
Implement Write Barrier

composerinteralia:
StringScanner holds the string being scanned, and a regex for methods
like `match?`. Triggering the write barrier for those allows us to mark
this as WB protected.

### DIFF
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -209,7 +209,7 @@ strscan_memsize(const void *ptr)
 static const rb_data_type_t strscanner_type = {
     "StringScanner",
     {strscan_mark, strscan_free, strscan_memsize},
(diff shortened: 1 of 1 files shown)


kou:
Do we also need to change https://github.com/ruby/strscan/blob/d4d9b10b3b5781dfb20e4a9d3121f67c0f76f4d0/ext/strscan/strscan.c#L63 ?

composerinteralia:
`nil` won't ever get collected, so I don't think there's anything special we need to do there. Even if we did this through `RB_OBJ_WRITE` I don't think that would change anything because we don't trigger the write barrier for special consts like `nil` (see https://github.com/ruby/ruby/blob/87d340f0e129ecf807e3be35d67fda1ad6f40389/include/ruby/internal/gc.h#L788-L789)

kou:
OK. Let's merge this.

PR #168, opened 2025-11-04
Remove methods have been obsolete over two decades

nobu:


rwstauner:
The `tzinfo` gem is using `rest?` which this PR appears to have removed from the C implementation but left in the Java implementation?
https://github.com/tzinfo/tzinfo/blob/v2.0.6/lib/tzinfo/data_sources/posix_time_zone_parser.rb#L61-L75

k0kubun:
> `rest?` which this PR appears to have removed from the C implementation but left in the Java implementation?

It wasn't marked "obsolete" in the first place, so I pushed a partial revert https://github.com/ruby/strscan/pull/169 to fix [a ruby/ruby CI](https://github.com/ruby/ruby/actions/runs/19080798257/job/54508537191).

PR #169, opened 2025-11-04
Resurrect a method that has not been obsolete

k0kubun:
Partially revert https://github.com/ruby/strscan/pull/168 because strscan_rest_p did not have `rb_warning("StringScanner#rest? is obsolete")`.

It is actively used by the latest tzinfo.gem, and we shouldn't remove it without deprecating it.


nobu:
That method is documented as obsolete since a6c2a8e4 at 2004-02-18.
Is it still used?

PR #186, opened 2025-12-31
retain code-range upon string extraction + align with strscan.c

kares:
added an extra step upon creating a substring from `scanner.string` -> retaining code-range when possible

also, there was an attempt to raise `"regexp buffer overflow"` upon `Matcher.INTERRUPTED` (-2), which isn't possible as the `RubyRegexp#matcherMatch/matcherSearch`  [handle the interrupt return](https://github.com/jruby/jruby/blob/10.0.2.0/core/src/main/java/org/jruby/RubyRegexp.java#L236-L257).

a follow-up on https://github.com/ruby/strscan/pull/185


headius (approved):

kares:
I don't have anything else atm, would be great to have a release.

PR #192, opened 2026-03-05
[Feature #21943] Add StringScanner#integer_at to extract capture group as Integer directly

jinroq:
see: https://bugs.ruby-lang.org/issues/21943

relation: https://bugs.ruby-lang.org/issues/21932

### DIFF
--- a/test/strscan/test_stringscanner.rb
+++ b/test/strscan/test_stringscanner.rb
@@ -1042,6 +1042,38 @@ module StringScannerTests
     assert_raise(ArgumentError) { s.integer_at(1) }
   end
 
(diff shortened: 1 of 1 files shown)


jinroq:
@kou 

> OK. Let's use integer_at(index).

Thanks, I renamed it to integer_at in [46a2e1d](https://github.com/ruby/strscan/pull/192/commits/46a2e1d2dbd77514a653453f9104e69ab6bd3fd2).

On a different note, I was trying to rebase but it seems I ended up implementing it on the master branch. Can I remake the PR or should I go ahead with it?

PR #193, opened 2026-03-06
[Feature #21943] Add `StringScanner#integer_at`

jinroq:
The specification for `MatchData#integer_at` has been defined [here](https://bugs.ruby-lang.org/issues/21932#note-7). `StringScanner#integer_at` follows this specification.

see: https://bugs.ruby-lang.org/issues/21932#note-6, https://bugs.ruby-lang.org/issues/21932#note-7
see: https://bugs.ruby-lang.org/issues/21943

### DIFF
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -1887,14 +1887,14 @@ parse_decimal_fast(const char *ptr, long len)
 
     /* Validate: only digits and underscores (not leading/trailing/consecutive) */
     {
-        long k;
+        long i;
         bool prev_underscore = true; /* treat start as underscore to reject leading _ */
-        for (k = digits_start; k < len; k++) {
-            if (ptr[k] >= '0' && ptr[k] <= '9') {
+        for (i = digits_start; i < len; i++) {
+            if (ptr[i] >= '0' && ptr[i] <= '9') {
                 digit_count++;
                 prev_underscore = false;
             }
-            else if (ptr[k] == '_' && !prev_underscore) {
+            else if (ptr[i] == '_' && !prev_underscore) {
                 prev_underscore = true;
             }
             else {
@@ -1911,19 +1911,19 @@ parse_decimal_fast(const char *ptr, long len)
     {
         long first_nonzero = digits_start;
         long effective_digits;
-        long k;
+        long i;
         while (first_nonzero < len && (ptr[first_nonzero] == '0' || ptr[first_nonzero] == '_'))
             first_nonzero++;
         effective_digits = 0;
-        for (k = first_nonzero; k < len; k++) {
-            if (ptr[k] != '_') effective_digits++;
+        for (i = first_nonzero; i < len; i++) {
+            if (ptr[i] != '_') effective_digits++;
         }
 
         if (effective_digits <= (sizeof(long) >= 8 ? INT64_DECIMAL_SAFE_DIGITS : INT32_DECIMAL_SAFE_DIGITS)) {
             long result = 0;
-            for (k = first_nonzero; k < len; k++) {
-                if (ptr[k] != '_')
-                    result = result * 10 + (ptr[k] - '0');
+            for (i = first_nonzero; i < len; i++) {
+                if (ptr[i] != '_')
+                    result = result * 10 + (ptr[i] - '0');
             }
             if (negative) result = -result;
             return LONG2NUM(result);
@@ -1935,9 +1935,9 @@ parse_decimal_fast(const char *ptr, long len)
                 ? (unsigned long)LONG_MAX + 1
                 : (unsigned long)LONG_MAX;
             bool overflow = false;
-            for (k = first_nonzero; k < len; k++) {
-                if (ptr[k] != '_') {
-                    unsigned long d = ptr[k] - '0';
+            for (i = first_nonzero; i < len; i++) {
+                if (ptr[i] != '_') {
+                    unsigned long d = ptr[i] - '0';
                     /* Pre-check before multiply to avoid unsigned long wraparound on
                      * 32-bit platforms, where 10-digit values can exceed ULONG_MAX. */
                     if (result > (limit - d) / 10) {
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -1919,19 +1919,19 @@ strscan_integer_at(int argc, VALUE *argv, VALUE self)
      * This covers the Date._strptime use case. */
     if (base == 10) {
         long j = 0;
-        int negative = 0;
+        bool negative = false;
         long digit_count;
 
-        if (ptr[0] == '-') { negative = 1; j = 1; }
+        if (ptr[0] == '-') { negative = true; j = 1; }
         else if (ptr[0] == '+') { j = 1; }
 
         digit_count = len - j;
         if (digit_count > 0) {
             long k;
-            int all_digits = 1;
+            bool all_digits = true;
             for (k = j; k < len; k++) {
                 if (ptr[k] < '0' || ptr[k] > '9') {
-                    all_digits = 0;
+                    all_digits = false;
                     break;
                 }
             }
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -1997,11 +1997,12 @@ strscan_integer_at(int argc, VALUE *argv, VALUE self)
     long i;
     long beg, end, len;
     const char *ptr;
-    VALUE specifier, vbase;
+    VALUE specifier;
     int base = 10;
 
-    rb_scan_args(argc, argv, "11", &specifier, &vbase);
-    if (!NIL_P(vbase)) base = NUM2INT(vbase);
+    rb_check_arity(argc, 1, 2);
+    specifier = argv[0];
+    if (argc > 1) base = NUM2INT(argv[1]);
 
     GET_SCANNER(self, p);
     i = resolve_capture_index(p, specifier);
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -1621,6 +1621,37 @@ name_to_backref_number(struct re_registers *regs, VALUE regexp, const char* name
                  rb_long2int(name_end - name), name);
 }
 
+/* Resolve capture group index from Integer, Symbol, or String.
+ * Returns the resolved register index, or -1 if unmatched/out of range. */
+static long
+resolve_capture_index(struct strscanner *p, VALUE idx)
+{
+    const char *name;
+    long i;
+
+    if (! MATCHED_P(p)) return -1;
+
+    switch (TYPE(idx)) {
+        case T_SYMBOL:
+            idx = rb_sym2str(idx);
+            /* fall through */
+        case T_STRING:
+            RSTRING_GETMEM(idx, name, i);
+            i = name_to_backref_number(&(p->regs), p->regex, name, name + i, rb_enc_get(idx));
+            break;
+        default:
+            i = NUM2LONG(idx);
+    }
+
+    if (i < 0)
+        i += p->regs.num_regs;
+    if (i < 0)                 return -1;
+    if (i >= p->regs.num_regs) return -1;
+    if (p->regs.beg[i] == -1)  return -1;
+
+    return i;
+}
+
 /*
  *
  * :markup: markdown
@@ -1695,30 +1726,12 @@ name_to_backref_number(struct re_registers *regs, VALUE regexp, const char* name
 static VALUE
 strscan_aref(VALUE self, VALUE idx)
 {
-    const char *name;
     struct strscanner *p;
     long i;
 
     GET_SCANNER(self, p);
-    if (! MATCHED_P(p))        return Qnil;
-
-    switch (TYPE(idx)) {
-        case T_SYMBOL:
-            idx = rb_sym2str(idx);
-            /* fall through */
-        case T_STRING:
-            RSTRING_GETMEM(idx, name, i);
-            i = name_to_backref_number(&(p->regs), p->regex, name, name + i, rb_enc_get(idx));
-            break;
-        default:
-            i = NUM2LONG(idx);
-    }
-
-    if (i < 0)
-        i += p->regs.num_regs;
-    if (i < 0)                 return Qnil;
-    if (i >= p->regs.num_regs) return Qnil;
-    if (p->regs.beg[i] == -1)  return Qnil;
+    i = resolve_capture_index(p, idx);
+    if (i < 0) return Qnil;
 
     return extract_range(p,
                          adjust_register_position(p, p->regs.beg[i]),
@@ -1850,6 +1863,60 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
     }
 
     return new_ary;
+}
+
+/*
+ * call-seq:
+ *   integer_at(index, base = 10) -> integer or nil
+ *
+ * Returns the captured substring at the given +index+ as an Integer,
+ * following the behavior of <tt>String#to_i(base)</tt>.
+ *
+ * +index+ can be an Integer (positive, negative, or zero), a Symbol,
+ * or a String for named capture groups.
+ *
+ * Returns +nil+ if:
+ * - No match has been performed or the last match failed
+ * - The +index+ is out of range
+ * - The group at +index+ did not participate in the match
+ *
+ * This is semantically equivalent to <tt>self[index].to_i(base)</tt>
+ * but avoids the allocation of a temporary String when possible.
+ *
+ *   scanner = StringScanner.new("2024-06-15")
+ *   scanner.scan(/(\d{4})-(\d{2})-(\d{2})/)
+ *   scanner.integer_at(1)       # => 2024
+ *   scanner.integer_at(1, 16)   # => 8228
+ *
+ */
+static VALUE
+strscan_integer_at(int argc, VALUE *argv, VALUE self)
+{
+    struct strscanner *p;
+    long i;
+    long beg, end, len;
+    const char *ptr;
+    VALUE idx, vbase;
+    int base = 10;
+
+    rb_scan_args(argc, argv, "11", &idx, &vbase);
+    if (!NIL_P(vbase)) base = NUM2INT(vbase);
+
+    GET_SCANNER(self, p);
+    i = resolve_capture_index(p, idx);
+    if (i < 0) return Qnil;
+
+    beg = adjust_register_position(p, p->regs.beg[i]);
+    end = adjust_register_position(p, p->regs.end[i]);
+    len = end - beg;
+
+    if (len <= 0) return INT2FIX(0);
+
+    ptr = S_PBEG(p) + beg;
+
+    /* Follow String#to_i(base) semantics via rb_str_to_inum.
+     * badcheck=0 returns 0 for non-numeric input instead of raising. */
+    return rb_str_to_inum(rb_str_new(ptr, len), base, 0);
 }
 
 /*
@@ -2290,6 +2357,7 @@ Init_strscan(void)
     rb_define_method(StringScanner, "size",        strscan_size,        0);
     rb_define_method(StringScanner, "captures",    strscan_captures,    0);
     rb_define_method(StringScanner, "values_at",   strscan_values_at,  -1);
+    rb_define_method(StringScanner, "integer_at",  strscan_integer_at, -1);
 
     rb_define_method(StringScanner, "rest",        strscan_rest,        0);
     rb_define_method(StringScanner, "rest_size",   strscan_rest_size,   0);
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -1867,20 +1867,20 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
 
 /*
  * call-seq:
- *   integer_at(index, base = 10) -> integer or nil
+ *   integer_at(specifier, base = 10) -> integer or nil
  *
- * Returns the captured substring at the given +index+ as an Integer,
+ * Returns the captured substring at the given +specifier+ as an Integer,
  * following the behavior of <tt>String#to_i(base)</tt>.
  *
- * +index+ can be an Integer (positive, negative, or zero), a Symbol,
+ * +specifier+ can be an Integer (positive, negative, or zero), a Symbol,
  * or a String for named capture groups.
  *
  * Returns +nil+ if:
  * - No match has been performed or the last match failed
- * - The +index+ is out of range
- * - The group at +index+ did not participate in the match
+ * - The +specifier+ is out of range
+ * - The group at +specifier+ did not participate in the match
  *
- * This is semantically equivalent to <tt>self[index].to_i(base)</tt>
+ * This is semantically equivalent to <tt>self[specifier].to_i(base)</tt>
  * but avoids the allocation of a temporary String when possible.
  *
  *   scanner = StringScanner.new("2024-06-15")
@@ -1896,14 +1896,14 @@ strscan_integer_at(int argc, VALUE *argv, VALUE self)
     long i;
     long beg, end, len;
     const char *ptr;
-    VALUE idx, vbase;
+    VALUE specifier, vbase;
     int base = 10;
 
-    rb_scan_args(argc, argv, "11", &idx, &vbase);
+    rb_scan_args(argc, argv, "11", &specifier, &vbase);
     if (!NIL_P(vbase)) base = NUM2INT(vbase);
 
     GET_SCANNER(self, p);
-    i = resolve_capture_index(p, idx);
+    i = resolve_capture_index(p, specifier);
     if (i < 0) return Qnil;
 
     beg = adjust_register_position(p, p->regs.beg[i]);
--- a/ext/strscan/strscan.c
+++ b/ext/strscan/strscan.c
@@ -1624,23 +1624,23 @@ name_to_backref_number(struct re_registers *regs, VALUE regexp, const char* name
 /* Resolve capture group index from Integer, Symbol, or String.
  * Returns the resolved register index, or -1 if unmatched/out of range. */
 static long
-resolve_capture_index(struct strscanner *p, VALUE idx)
+resolve_capture_index(struct strscanner *p, VALUE specifier)
 {
     const char *name;
     long i;
 
     if (! MATCHED_P(p)) return -1;
 
-    switch (TYPE(idx)) {
+    switch (TYPE(specifier)) {
         case T_SYMBOL:
-            idx = rb_sym2str(idx);
+            specifier = rb_sym2str(specifier);
             /* fall through */
         case T_STRING:
-            RSTRING_GETMEM(idx, name, i);
-            i = name_to_backref_number(&(p->regs), p->regex, name, name + i, rb_enc_get(idx));
+            RSTRING_GETMEM(specifier, name, i);
+            i = name_to_backref_number(&(p->regs), p->regex, name, name + i, rb_enc_get(specifier));
             break;
         default:
-            i = NUM2LONG(idx);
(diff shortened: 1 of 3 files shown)


ext/strscan/strscan.c
```diff
@@ -1852,6 +1852,112 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
     return new_ary;
 }
 
+#ifdef HAVE_RB_INT_PARSE_CSTR
+VALUE rb_int_parse_cstr(const char *str, ssize_t len, char **endp,
+                        size_t *ndigits, int base, int flags);
+#define RB_INT_PARSE_SIGN 0x01
```
kou: If https://github.com/ruby/ruby/pull/16322 is merged, this will report a duplicated definition warning.

ext/strscan/strscan.c
```diff
@@ -2290,6 +2396,7 @@ Init_strscan(void)
     rb_define_method(StringScanner, "size",        strscan_size,        0);
     rb_define_method(StringScanner, "captures",    strscan_captures,    0);
     rb_define_method(StringScanner, "values_at",   strscan_values_at,  -1);
+    rb_define_method(StringScanner, "integer_at",     strscan_integer_at,     1);
```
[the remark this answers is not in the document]
jinroq: [1630df8](https://github.com/ruby/strscan/pull/193/commits/1630df897a0f9ccb208e5c188698e938db1a4142) fixed it.

jinroq:
@kou 
[This comment](https://bugs.ruby-lang.org/issues/21932#note-7) finalizes the specification for `MatchData#integer_at(n)`. `StringScanner#integer_at(n)` has also been updated to comply with this specification. Please review it.

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,95 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+            long k;
+            int all_digits = 1;
+            for (k = j; k < len; k++) {
+                if (ptr[k] < '0' || ptr[k] > '9') {
+                    all_digits = 0;
+                    break;
+                }
+            }
+            if (all_digits) {
+                if (digit_count <= (sizeof(long) >= 8 ? 18 : 9)) {
```
kou: * It seems that `9223372036854775807` (max `int64_t` value) isn't optimized. Is it intentional?
* It seems that `00000000000000000001` isn't optimized. Is it intentional?

test/strscan/test_stringscanner.rb
```diff
@@ -968,6 +968,186 @@ def test_named_captures_same_name_union
     assert_equal({"number" => "1"}, scan.named_captures)
   end
 
+  def test_integer_at
```
kou: Could you use `test_integer_at_XXX` like other methods?

test/strscan/test_stringscanner.rb
```diff
@@ -968,6 +968,186 @@ def test_named_captures_same_name_union
...
+    assert_nil(s.integer_at(100))
+    assert_nil(s.integer_at(-3))
+  end
+
+  def test_integer_at_optional_group_not_matched
+    s = create_string_scanner("2024-06")
+    s.scan(/(\d{4})-(\d{2})(-(\d{2}))?/)
+    assert_equal(2024, s.integer_at(1))
+    assert_equal(6, s.integer_at(2))
+    assert_nil(s.integer_at(4))
```
kou: Is `4` correct?` If it's matched, it has a leading `-`. Is it intentional? Should we use `/...-(\d{2})?/` instead?

test/strscan/test_stringscanner.rb
```diff
@@ -968,6 +968,186 @@ def test_named_captures_same_name_union
...
+  end
+
+  def test_integer_at_fixnum_bignum_boundary
+    # 18 digits: fits in long on 64-bit
+    s = create_string_scanner("999999999999999999")
+    s.scan(/(\d+)/)
+    assert_equal(999999999999999999, s.integer_at(1))
+
+    # 19 digits: exceeds long on 64-bit, becomes bignum
+    s = create_string_scanner("9999999999999999999")
```
kou: In general, we should use border values for testing. If `"9" * 18` is the largest optimized value, we should use `"9" * 18"` and `"1" * 19"` (the next value of `"9" * 18"`).

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,95 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
     return new_ary;
 }
 
+/*
+ * call-seq:
+ *   integer_at(index, base = 10) -> integer or nil
```
kou: Could you use `specifier` not `index` like we did for `[]`?
https://github.com/ruby/strscan/blob/424375112caaada52c378c7bd354cdaf740cf0c0/ext/strscan/strscan.c#L1625-L1695

test/strscan/test_stringscanner.rb
```diff
@@ -968,6 +968,186 @@ def test_named_captures_same_name_union
...
+  end
+
+  def test_integer_at_leading_zeros
+    # "010" is 8 in octal (Integer("010")), but 10 in base 10
+    s = create_string_scanner("010")
+    s.scan(/(\d+)/)
+    assert_equal(10, s.integer_at(1))
+  end
+
+  def test_integer_at_full_match_with_non_digits
```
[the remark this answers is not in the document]
jinroq: [d89e54b](https://github.com/ruby/strscan/pull/193/commits/d89e54b96b3b1160d34b373e45dbadd88b149c05) fixed it.

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,147 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+                    : (unsigned long)LONG_MAX;
+                bool overflow = false;
+                for (; j < len; j++) {
+                    if (ptr[j] != '_') {
+                        unsigned long d = ptr[j] - '0';
+                        if (result > (limit - d) / 10) {
+                            overflow = true;
+                            break;
+                        }
+                        result = result * 10 + d;
```
kou: Can we avoid needless `(limit - d) / 10` computation?

```suggestion
                        result = result * 10 + d;
                        if (result > limit) {
                            overflow = true;
                            break;
                        }
```

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,147 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+                long result = 0;
+                for (; j < len; j++) {
+                    if (ptr[j] != '_')
+                        result = result * 10 + (ptr[j] - '0');
+                }
+                if (negative) result = -result;
+                return LONG2NUM(result);
+            }
+            /* One more digit than safe: may still fit in long with overflow check */
+            if (effective_digits <= (sizeof(long) >= 8 ? INT64_DECIMAL_SAFE_DIGITS + 1 : INT32_DECIMAL_SAFE_DIGITS + 1)) {
```
kou: Could you use `else if` for readability?

```suggestion
            }
            /* One more digit than safe: may still fit in long with overflow check */
            else if (effective_digits == (sizeof(long) >= 8 ? INT64_DECIMAL_SAFE_DIGITS + 1 : INT32_DECIMAL_SAFE_DIGITS + 1)) {
```

test/strscan/test_stringscanner.rb
```diff
@@ -968,6 +968,216 @@ def test_named_captures_same_name_union
...
+    assert_equal(large.to_i, s.integer_at(1))
+  end
+
+  def test_integer_at_fixnum_bignum_boundary
+    # 18 digits max ("9" * 18): largest value without overflow check
+    s = create_string_scanner("9" * 18)
+    s.scan(/(\d+)/)
+    assert_equal(("9" * 18).to_i, s.integer_at(1))
+
+    # 19 digits min ("1" * 19): smallest value with overflow check
```
kou: It seems that `"1" + "0" * 18` not `"1" * 19` is minimum?

kou (commented):
Could you also add a test when the current position isn't 0?

```ruby
scanner = StringScanner.new("a 10")
scanner.scan(/a /) # or scanner.pos = 2
scanner.scan(/\d+/)
scanner.integer_at(0)
```

ext/strscan/strscan.c
```diff
@@ -1622,6 +1621,37 @@ name_to_backref_number(struct re_registers *regs, VALUE regexp, const char* name
                  rb_long2int(name_end - name), name);
 }
 
+/* Resolve capture group index from Integer, Symbol, or String.
+ * Returns the resolved register index, or -1 if unmatched/out of range. */
+static long
+resolve_capture_index(struct strscanner *p, VALUE idx)
```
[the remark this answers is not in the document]
jinroq: [b6fe693](https://github.com/ruby/strscan/pull/193/commits/b6fe693818c6a2872fab774f56e0f1d7df66e42f)　fixed it.

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,147 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+                }
+                if (!overflow) {
+                    if (negative)
+                        return LONG2NUM(-(long)result);
+                    return LONG2NUM((long)result);
+                }
+            }
+            /* Bignum: fall through to rb_str_to_inum */
+        }
+    }
```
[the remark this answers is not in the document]
jinroq: [a078c0d](https://github.com/ruby/strscan/pull/193/commits/a078c0dc4ccfffbfdd47ed2bf70db906053a02f0) fixed it.

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,147 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+                            overflow = true;
+                            break;
+                        }
+                        result = result * 10 + d;
+                    }
+                }
+                if (!overflow) {
+                    if (negative)
+                        return LONG2NUM(-(long)result);
+                    return LONG2NUM((long)result);
```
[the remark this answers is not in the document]
jinroq: [b1b135a](https://github.com/ruby/strscan/pull/193/commits/b1b135ab96c87321ba295218fc9ef8dfe0ef9946) fixed it.

ext/strscan/strscan.c
jinroq: [db383e9](https://github.com/ruby/strscan/pull/193/commits/db383e9f92735b6abc06a7b8f30487c11a3e0364) fixed it.

ext/strscan/strscan.c
jinroq: [a421b12](https://github.com/ruby/strscan/pull/193/commits/a421b1228ed79b9c9b5391d08b1866032b1b93ed) fixed it.

test/strscan/test_stringscanner.rb
```diff
@@ -968,6 +968,216 @@ def test_named_captures_same_name_union
     assert_equal({"number" => "1"}, scan.named_captures)
   end
 
+  def test_integer_at_date_parts
```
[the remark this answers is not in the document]
jinroq: [a81aa3e](https://github.com/ruby/strscan/pull/193/commits/a81aa3e9ad432e814da7e9a879348f675219bb41) fixed it.

test/strscan/test_stringscanner.rb
jinroq: [d07b086](https://github.com/ruby/strscan/pull/193/commits/d07b0867adea0f674e3e8cedb9c368fefcf4a22c) fixed it.

test/strscan/test_stringscanner.rb
```diff
@@ -968,6 +968,216 @@ def test_named_captures_same_name_union
...
+    s.scan(/(\d+)/)
+    assert_equal(long_max + 1, s.integer_at(1))
+
+    # LONG_MIN - 1 (negative bignum)
+    s = create_string_scanner((long_min - 1).to_s)
+    s.scan(/([+\-]?\d+)/)
+    assert_equal(long_min - 1, s.integer_at(1))
+
+    # leading zeros with many digits
+    s = create_string_scanner("00000000000000000001")
```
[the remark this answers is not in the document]
jinroq: [1ae5772](https://github.com/ruby/strscan/pull/193/commits/1ae57726c300ed5749372ebffc1febb08d65178d) fixed it

test/strscan/test_stringscanner.rb
```diff
@@ -968,6 +968,216 @@ def test_named_captures_same_name_union
...
+  def test_integer_at_fixnum_bignum_boundary
+    # 18 digits max ("9" * 18): largest value without overflow check
+    s = create_string_scanner("9" * 18)
+    s.scan(/(\d+)/)
+    assert_equal(("9" * 18).to_i, s.integer_at(1))
+
+    # 19 digits min ("1" * 19): smallest value with overflow check
+    s = create_string_scanner("1" * 19)
+    s.scan(/(\d+)/)
+    assert_equal(("1" * 19).to_i, s.integer_at(1))
```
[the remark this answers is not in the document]
jinroq: The boundaries verified by these tests can be organized as follows:

| Test Value | Boundary Description |
| :--- | :--- |
| `"9" * 18` | Upper limit of the **safe path** (Max 18-digit value) |
| `"1" + "0" * 18` | Lower limit of the **overflow check path** (Min 19-digit value) |
| `long_max` | Upper limit within the **overflow check path** (Max value for `long`) |
| `long_max + 1` | Boundary for transitioning to **bignum** |

`long_max` verifies the "upper limit fitting in a `long` within the overflow check path," which is a distinct boundary from `"1" + "0" * 18`, which verifies the "transition from the safe path to the overflow check path (18 to 19 digits)."

Both test cases carry unique significance and should be retained.

Furthermore, with the [previous adjustment (`"1" * 19` -> `"1" + "0" * 18`)](https://github.com/ruby/strscan/pull/193/commits/d07b0867adea0f674e3e8cedb9c368fefcf4a22c), the test now correctly reflects the intended boundary value.

jinroq:
@kou 

> Could you also add a test when the current position isn't 0?

[6b44dbd](https://github.com/ruby/strscan/pull/193/commits/6b44dbd909d8d1ac5040b77a1fd4a03af2450931) fixed it.

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,166 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+ *
+ * This is semantically equivalent to <tt>self[specifier].to_i(base)</tt>
+ * but avoids the allocation of a temporary String when possible.
+ *
+ *   scanner = StringScanner.new("2024-06-15")
+ *   scanner.scan(/(\d{4})-(\d{2})-(\d{2})/)
+ *   scanner.integer_at(1)       # => 2024
+ *   scanner.integer_at(1, 16)   # => 8228
+ *
+ */
```
kou: Could you move this just before `static VALUE strscan_integer_at()`?

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,166 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+        while (first_nonzero < len && (ptr[first_nonzero] == '0' || ptr[first_nonzero] == '_'))
+            first_nonzero++;
+        effective_digits = 0;
+        for (k = first_nonzero; k < len; k++) {
+            if (ptr[k] != '_') effective_digits++;
+        }
+
+        if (effective_digits <= (sizeof(long) >= 8 ? INT64_DECIMAL_SAFE_DIGITS : INT32_DECIMAL_SAFE_DIGITS)) {
+            long result = 0;
+            for (; j < len; j++) {
```
kou: Can we use `for (k = first_nonzero; k < len; k++)` here?

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,166 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+            return LONG2NUM(result);
+        }
+        /* One more digit than safe: may still fit in long with overflow check */
+        else if (effective_digits == (sizeof(long) >= 8 ? INT64_DECIMAL_SAFE_DIGITS + 1 : INT32_DECIMAL_SAFE_DIGITS + 1)) {
+            unsigned long result = 0;
+            unsigned long limit = negative
+                ? (unsigned long)LONG_MAX + 1
+                : (unsigned long)LONG_MAX;
+            bool overflow = false;
+            for (; j < len; j++) {
```
kou: ditto.

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,166 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+#define INT32_DECIMAL_SAFE_DIGITS 9
+
+/* Fast path for base-10 integer parsing without temporary String allocation.
+ * Accepts digits and optional underscores (Ruby String#to_i semantics).
+ * Returns a Fixnum/Integer VALUE on success, or Qundef to signal fall-through
+ * to the general path (non-decimal, bignum, or non-numeric input). */
+static inline VALUE
+parse_decimal_fast(const char *ptr, long len)
+{
+    long j = 0;
```
kou: Can we use meaningful name instead of `j`? I think that this is not a loop variable.

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,166 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+    bool negative = false;
+    long digit_count = 0;
+    bool valid = true;
+
+    if (ptr[0] == '-') { negative = true; j = 1; }
+    else if (ptr[0] == '+') { j = 1; }
+
+    /* Validate: only digits and underscores (not leading/trailing/consecutive) */
+    {
+        long k;
```
kou: Can we use `i` not `k`?

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,166 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+                    result = result * 10 + d;
+                    if (result > limit) {
+                        overflow = true;
+                        break;
+                    }
+                }
+            }
+            if (!overflow) {
+                if (negative) {
+                    if (result == (unsigned long)LONG_MAX + 1)
```
kou: Can we use `limit` here?

lib/strscan/strscan.rb
```diff
@@ -22,4 +22,14 @@ def scan_integer(base: 10)
...
     end
   end
+
+  unless method_defined?(:integer_at)
+    # Fallback implementation for platforms without C extension (e.g. JRuby).
+    # Equivalent to self[specifier].to_i(base).
+    def integer_at(specifier, base = 10)
+      str = self[specifier]
+      return nil if str.nil?
+      str.to_i(base)
```
kou: Can we use `self[specifier]&.to_i(base)`?

ext/strscan/strscan.c
```diff
@@ -1852,6 +1865,166 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+                : (unsigned long)LONG_MAX;
+            bool overflow = false;
+            for (; j < len; j++) {
+                if (ptr[j] != '_') {
+                    unsigned long d = ptr[j] - '0';
+                    result = result * 10 + d;
+                    if (result > limit) {
+                        overflow = true;
+                        break;
+                    }
```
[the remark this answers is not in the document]
jinroq: [e6e5c27](https://github.com/ruby/strscan/pull/193/commits/e6e5c27fa127e51950c44895e9cb2640453834ed) fixed it

ext/strscan/strscan.c
```diff
@@ -1852,6 +1865,166 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+{
+    struct strscanner *p;
+    long i;
+    long beg, end, len;
+    const char *ptr;
+    VALUE specifier, vbase;
+    int base = 10;
+
+    rb_scan_args(argc, argv, "11", &specifier, &vbase);
+    if (!NIL_P(vbase)) base = NUM2INT(vbase);
```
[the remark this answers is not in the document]
jinroq: [ba15508](https://github.com/ruby/strscan/pull/193/commits/ba15508f1cb7d99fb1d1d2e28dc40be43049f276) fixed it.

test/strscan/test_stringscanner.rb
```diff
@@ -968,6 +968,225 @@ def test_named_captures_same_name_union
...
+  end
+
+  def test_integer_at_base_zero
+    # base 0 respects prefixes like 0x
+    s = create_string_scanner("0xF")
+    s.scan(/(...)/)
+    assert_equal(0, s.integer_at(1))       # base 10: "0xF".to_i => 0
+    assert_equal(15, s.integer_at(1, 0))   # base 0: "0xF".to_i(0) => 15
+  end
+
```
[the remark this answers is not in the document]
jinroq: https://github.com/ruby/strscan/pull/193#discussion_r3021882339 fixed it.

ext/strscan/strscan.c
jinroq: This fix was reverted to [e6e5c27](https://github.com/ruby/strscan/pull/193/commits/e6e5c27fa127e51950c44895e9cb2640453834ed) because it did not take 32-bit environments into consideration.

A comment was added in [d5b3651](https://github.com/ruby/strscan/pull/193/commits/d5b3651766bfca36894490a8c7795208fa7e63cd).

ext/strscan/strscan.c
```diff
@@ -1853,6 +1865,169 @@ strscan_values_at(int argc, VALUE *argv, VALUE self)
...
+        if (effective_digits <= (sizeof(long) >= 8 ? INT64_DECIMAL_SAFE_DIGITS : INT32_DECIMAL_SAFE_DIGITS)) {
+            long result = 0;
+            for (; j < len; j++) {
+                if (ptr[j] != '_')
+                    result = result * 10 + (ptr[j] - '0');
+            }
+            if (negative) result = -result;
+            return LONG2NUM(result);
+        }
+        /* One more digit than safe: may still fit in long with overflow check */
```
tompng, line 1931: What do you think of combining validation/counting/safe_digits_branch/one_more_digit_branch to a single for-loop like this?
```c
for (k = j; k < len; k++) { //single moderate-fast path
  if (invalid) stop_parsing_or_fallback_to_slow_path;
  if (result > (limit - d) / 10) fallback_to_slow_path; // always check overflow
  result = result * 10 + d;
}
```
It always checks overflow, which is an overhead. It always performs multiplication even for `> 19` digits case.
But code is simple, less branch, boundary test will be simple too.

Compared to this, the current code is really optimized to remove all the overhead that the simple for-loop has.
- validation: to eliminate multiplication for invalid case
- counting: to eliminate useless multiplication for large number of digits case
- safe_digits_branch: fast path to eliminate overflow check

I just want to know if the performance improvement of these optimizations are really worth adding complexity to the code.
My guess is, a simple and moderate fast path is fast enough for the use case.

If the overflow check overhead needs to be reduced, something like `result > (LONG_MAX-9)/10 && result > (limit-d)/10` might be performant.

ext/strscan/strscan.c
jinroq: [9efb4d2](https://github.com/ruby/strscan/pull/193/commits/9efb4d2f2aea871a808f3d1fc05c0989fa6e458a) fixed it.

ext/strscan/strscan.c
jinroq: [f51e1cf](https://github.com/ruby/strscan/pull/193/commits/f51e1cfa34b15adb4accf3a51d78853f53eb0bcb) fixed it.

ext/strscan/strscan.c
jinroq: [f51e1cf](https://github.com/ruby/strscan/pull/193/commits/f51e1cfa34b15adb4accf3a51d78853f53eb0bcb) fixed it.

ext/strscan/strscan.c
jinroq: [764a4a1](https://github.com/ruby/strscan/pull/193/commits/764a4a1a9b1e9e4e10b14bca3ec4416c906985ce) fixed it.

ext/strscan/strscan.c
jinroq: [0f67657](https://github.com/ruby/strscan/pull/193/commits/0f6765713f5799c802c3e4a55def995164db1b4f) fixed it.

ext/strscan/strscan.c
jinroq: [7f19293](https://github.com/ruby/strscan/pull/193/commits/7f19293a98663eeacdbe0c8220d33841200ff6e4) fixed it.

lib/strscan/strscan.rb
jinroq: [e4c6a1e](https://github.com/ruby/strscan/pull/193/commits/e4c6a1ef0994813bb6f5453e5399ea1960a09211) fixed it.

jinroq:
@kou 

I’d like to hear your thoughts on [this comment](https://github.com/ruby/strscan/pull/193#discussion_r3033996293).

Regarding `Date._strptime`, this comment makes a valid point.
Since the year is 4 digits, the month is 2 digits, and the day is 2 digits - totaling 18 digits or fewer - paths other than the safe path (such as the counting and overflow check paths) are rarely reached in actual use cases.
Compared to the two-pass approach currently used in the code (validation + counting -> accumulation), a simple single loop can be expected to achieve comparable performance in most cases.

However, `StringScanner#integer_at` is not a feature intended solely for `Date`.

- Since a `limit` must be set for inputs with more than 19 digits (as in the `test_integer_at_large_number` case), the `limit` must be determined before the `negative` is finalized, which can ultimately result in code that is at least as complex
- The current implementation correctly handles all cases involving signs, underscores, and leading zeros; if replaced with a simple loop, these operations would need to be crammed into a single loop.

I don’t think I need to respond.

kou:
Could you add your comment to the thread https://github.com/ruby/strscan/pull/193#discussion_r3033996293 for easy to follow?

ext/strscan/strscan.c
jinroq: @kou 

I’d like to hear your thoughts on [this comment](https://github.com/ruby/strscan/pull/193#discussion_r3033996293).

Regarding `Date._strptime`, this comment makes a valid point.
Since the year is 4 digits, the month is 2 digits, and the day is 2 digits - totaling 18 digits or fewer - paths other than the safe path (such as the counting and overflow check paths) are rarely reached in actual use cases.
Compared to the two-pass approach currently used in the code (validation + counting -> accumulation), a simple single loop can be expected to achieve comparable performance in most cases.

However, `StringScanner#integer_at` is not a feature intended solely for `Date`.

- Since a `limit` must be set for inputs with more than 19 digits (as in the `test_integer_at_large_number` case), the `limit` must be determined before the `negative` is finalized, which can ultimately result in code that is at least as complex
- The current implementation correctly handles all cases involving signs, underscores, and leading zeros; if replaced with a simple loop, these operations would need to be crammed into a single loop.

I don’t think I need to respond.

jinroq:
@kou 
Sorry, I mistook this for a regular comment.
I left a comment [here](https://github.com/ruby/strscan/pull/193#discussion_r3037553517).

ext/strscan/strscan.c
kou: Let's consider with numbers.
Could you benchmark these 2 approaches? If there is no significant performance difference, the simpler approach is better for maintenance.

BTW, why did you close https://github.com/ruby/ruby/pull/16322 ? If `rb_int_parse_cstr()` is exported, we don't need to implement this, right?

kou: @jinroq @tompng How about #205? If we need more performance and our custom `rb_int_parse_cstr()` implementation like we do in this PR is faster, we can work on it later.

kou: It seems that there is no objection. I'll merge #205 in a few days.

kou:
I close this in favor of #205 .

PR #203, opened 2026-04-27
Fix `call-seq` formatting

Earlopain:
They don't work in included markdown and also need a leading `:`

grafik


kou:
@st0012 Is it an expected RDoc behavior that `call-seq` doesn't work in `include`?

kou:
It's an expected behavior in the current RDoc implementation but `call-seq` in `include` will be an useful feature.
Could you open a feature request issue to https://github.com/ruby/rdoc for `call-seq` in `include`?

PR #207, opened 2026-05-21
Fix call-seq return values

zonuexe:
Some `call-seq` comments did not match the actual return values (documentation only, no behavior change):

* `match?` returns the match size, not a position
* `skip` was missing the `->` arrow
* `scan_full` can also return a length
* `scan_byte` can return `nil` at end of string


kou:
Good catch!

PR #216, opened 2026-08-06
Remove obsolete constants

nobu:
Undocumented `::ScanError` and `::Id`.  `StringScanner::ScanError` should be used instead of the former, and the latter is completely useless now.


eregon (approved):
Nice cleanup!
IIRC there might be a test or spec checking that, but regardless that test/spec should be removed.

PR #217, opened 2026-09-10
Fix matched_size when the stored string is shrunk

youdie006:
`matched_size` returns the raw register width without clamping it to the current length of the stored string, so shrinking the string leaves it reporting the old match:

```ruby
s = StringScanner.new("abcdef")
s.scan(/abcdef/)
s.string = "abc"
s.matched_size   # => 6, should be 3
s.matched        # => "abc"   (this one is already correct)
```

`extract_range` (`ext/strscan/strscan.c:168-169`) already clamps, which is why `matched` answers correctly. #212 added the same two lines to `integer_at`, `charpos` got them in `7b77f30`, and `bol?` carries its own guard — `matched_size` was missed.

Fixed in all three backends (CRuby, JRuby, TruffleRuby) with a shared regression test.


test/strscan/test_stringscanner.rb
```diff
@@ -738,6 +738,27 @@ def test_matched_size
...
+
+    s.string.replace("before 2")
+    assert_equal("2", s.matched)
+    assert_equal(1, s.matched_size)
+
+    s.string.replace("before ")
+    assert_equal("", s.matched)
+    assert_equal(0, s.matched_size)
+
+    s.string.replace("bef")
```
kou, line 757: Could you use `before` as a boundary value??

```suggestion
    s.string.replace("before")
```

kou (approved):
Could you simplify the PR description?

youdie006:
Simplified — thanks, and sorry for the wall of text.

youdie006:
The truffleruby rows were red, and the cause turned out to be worth widening the PR slightly.

`lib/strscan/truffleruby.rb` had `matched = @last_match&.to_s`, which re-reads nothing and so returns the original match text after the string shrinks. Fixing only `matched_size` there made the two disagree with each other — `matched_size` would report 1 while `matched` still returned `"29"`. So `matched` now clamps the same way, mirroring `extract_range`:

```ruby
beg = Primitive.match_data_byte_begin(@last_match, 0)
return if beg > @string.bytesize
fin = [Primitive.match_data_byte_end(@last_match, 0), @string.bytesize].min
@string.byteslice(beg, fin - beg)
```

All four backends now use the same shape as `extract_range` (`ext/strscan/strscan.c:168-169`): `nil` when the match begins past the end, otherwise the end clamped to the current length.

CRuby is green locally — 167 tests, 1114 assertions, 0 failures.

One thing I could not verify locally, so I would rather say it than imply otherwise: `lib/strscan.rb:12` only loads `truffleruby.rb` when `RUBY_ENGINE_VERSION.to_i >= 34`, and the TruffleRuby image I have is 25.0.0, so it takes the stdlib branch. I confirmed from the failing job logs that CI runs 34.0.1 and 40.0.0 on those rows, so they do exercise this file — but the truffleruby rows here are verified by CI, not by me.

test/strscan/test_stringscanner.rb
kou: Let's try this.

kou (approved):
+1

We may want to change `[]`, `pre_match` and so on in the TruffleRuby implementation for shrunk case as a separated tasks.

PR #219, opened 2026-09-20
Don't match from the middle of a character

kou:
`#pos=` accepts an arbitrary byte offset, so the scan position can fall in the middle of a multi-byte character such as UTF-16BE. Matching from there made Onigmo read a whole character starting at that position, reading past the end of the string.

On CRuby this is an out-of-bounds read.

On JRuby a search such as `#scan_until` could loop without terminating.

On TruffleRuby it aborts with an internal error while building the matched substring.

Refuse to match unless the current position is the head of a character.


lib/strscan/truffleruby.rb
```diff
@@ -239,12 +239,30 @@ def search_full(pattern, advance_pointer, return_string)
   end
   Primitive.always_split self, :search_full
 
+  def pos_at_char_boundary?
+    pos = @pos
+    return true if pos == 0 || pos == @string.bytesize
+    return true if @string.bytesize == @string.length
+    begin
+      @string.b.byteslice(0, pos).force_encoding(@string.encoding).valid_encoding?
```
kou: @eregon This implementation is slow. Is there any better way to detect character boundary in TruffleRuby?

eregon: Could we validate on `pos=` instead maybe? That would avoid adding overhead on the fast path.

This seems indeed very slow, and I think between the existing exception and this expensive check, the existing exception is better, because users are very unlikely to hit it.
So I think no changes on TruffleRuby + skipping that test is better.

The exception is fatal though as we can see [here](https://github.com/ruby/strscan/actions/runs/35508979284/job/106073479240#step:8:77), that's something we should fix in TruffleRuby to be some Ruby exception instead: https://github.com/truffleruby/truffleruby/issues/4474

eregon: (FWIW there is `Primitive.string_is_character_head?` but that's a private Primitive so it cannot be used outside TruffleRuby's core library.)

eregon:
Returning `nil` might be confusing for the user as `nil` means "no match".
Maybe an exception would be better?
If we can move this validation to `pos=` I think that'd be even better.

lib/strscan/truffleruby.rb
kou: If we move the validation to `pos=`, we can't avoid other cases such as `get_byte`.

I'll not change the TruffleRuby implementation in this PR.

kou:
> Returning `nil` might be confusing for the user as `nil` means "no match". Maybe an exception would be better?

This case (match from the middle of a character) must be a "no match" case. This is a fast path for this case. (This also avoids false-positive match by `s = StringScanner.new("ab".encode("UTF-16BE"); s.pos = 1; s.scan("\x61\x00".force_encoding("UTF-16BE"))`.)

lib/strscan/truffleruby.rb
eregon: Thanks, I think not changing TruffleRuby is the best trade-off in this PR.
I'll consider exposing `Primitive.string_is_character_head?` as a public Primitive for the next release.

kou: OK. Could you work on it as a follow-up PR?
I'll merge this PR as-is.

